PR #12, opened 2017-02-08
genheaders.py: make it usable as module

groleo:
Signed-off-by: Adrian Negreanu <groleo@gmail.com>

### DIFF
--- a/xml/genheaders.py
+++ b/xml/genheaders.py
@@ -37,40 +37,6 @@ errFilename = None
 diagFilename = 'diag.txt'
 regFilename = 'gl.xml'
 
-if __name__ == '__main__':
-    i = 1
-    while (i < len(sys.argv)):
-        arg = sys.argv[i]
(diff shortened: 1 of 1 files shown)


oddhack:
I'm not particularly opposed to this but am curious about the use case. Does it accomplish something other than making it possible to import the data structures with the various options for different headers to another module? Really this whole thing is just a giant frontend wrapper designed to collect all those ni one place for relatively easy maintenance.

I'll check to make sure it doesn't affect the Makefile targets, but otherwise, seems fine.

groleo:
The way I'm using it is to have a genshim.py.
I import genheaders.py for its buildList and defs.
I then add a genShim() method and modify each filename from buildList[]
so that it outputs .cpp files instead of headers.

pdaniell-nv:
@oddhack if you're happy with this it's fine with me to merge.

PR #20, opened 2017-02-15
Fix bug number typo in EXT_YUV_target change log.

nvpbrown:


nvpbrown:
I had some questions for Jeff Leger when I was diffing versions of this spec, and he pointed me at the correct bug motivating his edits.

PR #23, opened 2017-02-21
Make GL_EXT_direct_state_access available in core profile

pdaniell-nv:
This is needed for an upcoming CTS test which relies on
glTexturePageCommitmentEXT, which should be available in both
compatibility and core profiles.


oddhack:
OK, but shouldn't the EXT_dsa spec say something about profile support as well?

Also, I think this will result in changes in some of the generated headers. Will take a look at that shortly.

oddhack:
Yeah, glcorearb.h changes as expected. Pushing now.

PR #31, opened 2017-02-27
Add EXT_multisampled_render_to_texture2

jackohound:
Adds EXT_multisampled_render_to_texture2 extension.
Discussion of the extension was tracked in Khronos-internal Bugzilla [16022](https://cvs.khronos.org/bugzilla/show_bug.cgi?id=16022).

### DIFF
--- a/extensions/EXT/EXT_multisampled_render_to_texture2.txt
+++ b/extensions/EXT/EXT_multisampled_render_to_texture2.txt
@@ -1,0 +1,109 @@
+Name
+
+    EXT_multisampled_render_to_texture2
+
+Name Strings
+
+    GL_EXT_multisampled_render_to_texture2
+
+Contributors
+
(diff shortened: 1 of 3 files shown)


extensions/EXT/EXT_multisampled_render_to_texture2.txt
```diff
@@ -0,0 +1,109 @@
...
+    Complete
+
+Version
+
+    Last Modified Date: February 27, 2017
+    Revision: 3
+
+Number
+
+    274
```
oddhack: Should be 275; I'll fix this after merging.

oddhack: Also, for future reference, should be "OpenGL ES Extension #275" not just "275" in the extension spec.

PR #40, opened 2017-03-25
Whitelist a number of ARB/NV extensions for glcore

dgkoch:
Add 'glcore' to the supported attribute for a good number of ARB/NV extensions that can definitely be supported in the core profile.  I'm sure there are a lot more that could be yet, but it's not entirely clear what the criteria is. 

A couple of other tweaks for extensions that had incorrect supported fields as well.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -37448,7 +37448,7 @@ typedef unsigned int GLhandleARB;
                 <command name="glMemoryBarrierByRegion"/>
             </require>
         </extension>
-        <extension name="GL_ARB_ES3_2_compatibility" supported="gl">
+        <extension name="GL_ARB_ES3_2_compatibility" supported="gl|glcore">
             <require>
                 <enum name="GL_PRIMITIVE_BOUNDING_BOX_ARB"/>
                 <enum name="GL_MULTISAMPLE_LINE_WIDTH_RANGE_ARB"/>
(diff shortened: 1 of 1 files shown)


xml/gl.xml
```diff
@@ -38703,7 +38703,7 @@ typedef unsigned int GLhandleARB;
                 <enum name="GL_ANY_SAMPLES_PASSED"/>
             </require>
         </extension>
-        <extension name="GL_ARB_parallel_shader_compile" supported="gl">
+        <extension name="GL_ARB_parallel_shader_compile" supported="gl|glcore">
```
dgkoch: This one is needed so the CTS loads the entry points for this extension.

xml/gl.xml
```diff
@@ -39287,9 +39287,9 @@ typedef unsigned int GLhandleARB;
                 <command name="glTexPageCommitmentARB"/>
             </require>
         </extension>
-        <extension name="GL_ARB_sparse_texture2" supported="gl|glcore|gles2"/>
```
dgkoch: ARB extensions should never be listed for gles2

xml/gl.xml
```diff
@@ -42242,7 +42242,7 @@ typedef unsigned int GLhandleARB;
                 <!-- <command name="glTexturePageCommitmentEXT"/> -->
             </require>
         </extension>
-        <extension name="GL_EXT_sparse_texture2" supported="gl"/>
+        <extension name="GL_EXT_sparse_texture2" supported="gl|gles2"/>
```
dgkoch: this one is registered for ES

oddhack: I didn't have any rigorous process for this. I would kinda glance at the extension spec and go off what it said about required GL versions, or otherwise guess. The only thing it drives in this repo is what goes in glcorearb.h, which I suspect is barely used by anyone, but happy to be more accurate.

dgkoch: The VK-GL-CTS drives what function pointers are loaded for the required functions off of this as well. For GL entry points it only loads them if they have 'glcore' specified.

PR #54, opened 2017-05-11
add EXT_EGL_image_array

jackohound:
I'd like to merge these two commits:
- Publish extension as agreed in [internal MR3](https://gitlab.khronos.org/opengl/WG-Docs/merge_requests/3).
- Minor updates to vendor extension QCOM_framebuffer_foveated

I pushed both commits onto my master branch.  Wasn't sure how to split these two commits into separate pull requests.     Let me know if you want these in separate merge requests.


oddhack:
We are also asking people to add their extension API to gl.xml so it gets included in the headers, which doesn't seem to be present in this MR. Can you make that update?

jackohound:
This extension does not add any new entrypoints or add any new tokens.  It just allows an existing token to be used with an existing entrypoint.  So no changes to gl.xml (or the headers)  are expected AFAIK.  

Was there some particular change you were expecting in gl.xml ?

oddhack:
Even if it has no API, we still describe the extensions in gl.xml with a tag like this:

>         <extension name="GL_SGIX_ycrcb_subsample" supported="gl|glcore"/>

(alphabetized in the \<extensions> block). Or supported='gl|glcore|gles2' if it's for both desktop and ES.

jackohound:
@oddhack, thanks for the pointers.  I've reworked the change to include the update to gl.xml.   (I'll submit the updates to updates QCOM_framebuffer_foveated as a separate pull request.)

PR #55, opened 2017-05-11
Add ARB_dsa interaction to GL_ARB_sparse_texture

nhaehnle:
Bugzilla: https://cvs.khronos.org/bugzilla/show_bug.cgi?id=13684


pdaniell-nv:
@nhaehnle Your change looks good. @oddhack Is there an XML change that needs to go along with this? If not, please merge. Thanks.

oddhack:
@pdaniell-nv there's already tagging in gl.xml to note that TexturePageCommitmentEXT requires sparse_texture as well as dsa. That doesn't affect the generated header file (nor should it), but at least the information is present.

PR #61, opened 2017-05-23
Assign extension number to GL_AMD_texture_gather_bias_lod + registry updates

DominikWitczakAMD:


extensions/AMD/AMD_texture_gather_bias_lod.txt
```diff
@@ -32,7 +32,7 @@ Version
 
 Number
 
-    TBD
+    502
```
oddhack: Should say "OpenGL Extension #502". But that can be fixed later.

PR #64, opened 2017-05-31
Add {GLX,WGL}_ARB_create_context_no_error to the registry

pdaniell-nv:
Hopefully everything is correct here.


oddhack:
Just one minor change: instead of using duplicate entries for the GLX and WGL extensions in the same file, use the 'alias' key in registry.py. Otherwise, looks great.

PR #65, opened 2017-06-06
Publish the GL/{Vulkan,D3D} interop extensions

cubanismo:
This commit publishes the following extensions:

  GL_EXT_memory_object
  GL_EXT_memory_object_fd
  GL_EXT_memory_object_win32
  GL_EXT_semaphore
  GL_EXT_semaphore_fd
  GL_EXT_semaphore_win32
  GL_EXT_win32_keyed_mutex

Which allow sharing textures, buffer objects, and
synchronization primitives with Vulkan, D3D, and
potentially other APIs in the future.


oddhack:
It appears that you've got this branch setup so that it's imposed Unix newline endings on specs/gl/glx1.3.pdf . I'm not entirely sure what the deal is here - there are plenty of other PDFs with both types of line endings (as seen in my local clone of the KhronosGroup repo, anyway).  I doubt this is harmful, though I'm also hoping it doesn't setup churn between versions in future commits.

Anyway, otherwise this looks fine, but you might look at whatever Git attribute settings are relevant to see why the DOS->Unix change happened on this file. Merging.

cubanismo:
Thanks Jon.  I'm not sure what happened there.  I didn't touch those files in any commits so I'm not clear why they would have had their line endings modified even if I did have git misconfigured somehow.  I wonder if it's a product of creating the fork in github somehow.

PR #69, opened 2017-06-12
Clarify interactions between EXT_protected_textures and EXT_sparse_te…

jackohound:
As agreed in Khronos internal [bug#16211](https://cvs.khronos.org/bugzilla/show_bug.cgi?id=16211).


TobiasHector:
Merging since text was already agreed upon.

PR #73, opened 2017-06-23
Fix typo in MemoryObjectParameterivEXT

lostgoat:
From the EXT_external_objects extension


lostgoat:
I'm not sure if I also need to update the extension revision for this type of change. Let me know if I should.

PR #76, opened 2017-06-30
Add types to GLES/gl.h for backwards compatibility

scott-d-phillips:
byte, clampf, short, and ushort are not used by the API but were
historically provided by GLES/gl.h

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -34791,6 +34791,12 @@ typedef unsigned int GLhandleARB;
 
     <!-- SECTION: OpenGL ES 1.0/1.1 API interface definitions. -->
     <feature api="gles1" name="GL_VERSION_ES_CM_1_0" number="1.0">
+        <require comment="Not used by the API, for compatibility with old gl.h">
+            <type name="GLbyte"/>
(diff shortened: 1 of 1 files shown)


scott-d-phillips:
Fixes: #77

oddhack:
Can you update the PR without the generated headers? That invariably generates conflicts when other PRs go in affecting the same headers, so it's easiest to re-do it myself.

scott-d-phillips:
@oddhack Makes sense, updated.

1ace:
Ping? We just got a [bug report](https://bugs.freedesktop.org/show_bug.cgi?id=102003) caused by this, it'd be great if this gets merged so we can update the headers in Mesa, especially since a [release](https://mesa3d.org/release-calendar.html) is coming up :)

oddhack:
@1ace I need to get signoff from the ES WG before accepting this. I don't think anyone's likely to object, it's just a matter of getting some of their cycle time to upvote it.

TobiasHector:
Telecon 2017-08-09: ES WG agreed to take this change - whenever you're happy with the PR Jon.

PR #80, opened 2017-07-25
Clarify Blit and Clear behaviours, and fixed framebuffer attachment binding psuedocode

TobiasHector:
Fixes internal bugzillas 16173, 16174, 16176

CC @casseveritt, since you're the contact on these.


TobiasHector:
The OpenGL ES WG signed off this change today, so accepting into the registry.

PR #86, opened 2017-07-28
Fix extension reference in EXT_EGL_image_array

jackohound:
Fix reference to extension as agreed in Khronos-internal  [WG-Docs Issue#6](https://gitlab.khronos.org/opengl/WG-Docs/issues/6)


dgkoch:
+1

PR #88, opened 2017-07-30
Add EXT_texture_compression_s3tc_srgb for ES

Groovounet:


Groovounet:
I didn't know how to run the tools to produce the generated files. Is there a doc somewhere?

Thanks,
Christophe

oddhack:
Yes, the toplevel README. You will also need to fix the conflict.

extensions/registry.py
```diff
@@ -2314,6 +2314,12 @@
         'supporters' : { 'INTEL', 'NVIDIA' },
         'url' : 'extensions/EXT/EXT_texture_compression_s3tc.txt',
     },
+    'GL_EXT_texture_compression_s3tc_srgb' : {
+        'esnumber' : 286,
```
dgkoch: I believe the next available ES number is 289 (please run 'python nextfree.py' in the extensions folder to find the next available number.

Groovounet: Fixed, my clone was behind in number of extensions registred :D

oddhack:
I've updated the gl2ext.h header with this extension now as well.

Groovounet:
Thanks!

PR #97, opened 2017-08-17
Add missing include of stddef.h for ptrdiff_t in GL_ARB_vertex_buffer_object

jeremyhu:
Fixes https://www.khronos.org/bugzilla/show_bug.cgi?id=1345

Signed-off-by: Jeremy Huddleston Sequoia <jeremyhu@apple.com>

### DIFF
--- a/api/GL/glext.h
+++ b/api/GL/glext.h
@@ -4713,6 +4713,7 @@ GLAPI void APIENTRY glVertexBlendARB (GLint count);
 
 #ifndef GL_ARB_vertex_buffer_object
 #define GL_ARB_vertex_buffer_object 1
+#include <stddef.h>
 typedef ptrdiff_t GLsizeiptrARB;
 typedef ptrdiff_t GLintptrARB;
 #define GL_BUFFER_SIZE_ARB                0x8764


oddhack:
TL;DR while I don't yet understand what you're trying to accomplish, this isn't a good way to go about it. If you will explain the usage scenario and make clear why having the stddef.h include elsewhere in the header is a problem, we may be able to find a better answer for you. But the structure of the header files, and the fact they're generated from gl.xml, doesn't support what you're asking for. This PR would be instantly overridden the next time the headers are generated.

I am closing the bugzilla https://www.khronos.org/bugzilla/show_bug.cgi?id=1345 as we're trying to migrate completely off that platform.

jeremyhu:
Currently, `<stddef.h>` is only included by glext.h if `GL_VERSION_1_5` isn't defined.

So if `GL_VERSION_1_5` is defined and `GL_ARB_vertex_buffer_object` is not defined when glext.h is included, we will encounter a parse error.

This change addresses that problem.

oddhack:
Yes, I understand that. What I don't understand is why it's causing you problems. The only time GL_VERSION_1_5, or any other reserved version or extension symbol in the header, should be defined in the header is if, for some arcane reason, you are *replacing* the definition of those interfaces with something equivalent prior to including the header. So what is the specific use case under which this is happening for you, and why isn't that replacement code itself defining ptrdiff_t?

jeremyhu:
On macOS, CGL's gl.h defines GL_VERSION_1_5 but not GL_ARB_vertex_buffer_object.

Thus, including OpenGL.framework's gl.h followed by the current glext.h will result in a failure.  See:

/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.Internal.sdk/System/Library/Frameworks/OpenGL.framework/Headers $ egrep '(GL_VERSION_1_5|GL_ARB_vertex_buffer_object)' *h
CGLMacro.h:/* GL_ARB_vertex_buffer_object */
gl.h:#define GL_VERSION_1_5                    1
gl3.h:#ifndef GL_VERSION_1_5
gl3.h:#ifndef GL_VERSION_1_5
gl3.h:#define GL_VERSION_1_5 1
glext.h:#define GL_ARB_vertex_buffer_object                          1
glext.h:#if GL_ARB_vertex_buffer_object
glext.h:#if GL_ARB_vertex_buffer_object
glext.h:#if GL_ARB_vertex_buffer_object
gltypes.h:#if !defined(GL_VERSION_1_5)
gltypes.h:#if !defined(GL_ARB_vertex_buffer_object)

PR #99, opened 2017-09-05
Added EXT_texture_mirror_clamp_to_edge extension

Groovounet:
OpenGL ES is missing an extension to expose MIRROR_CLAMP_TO_EDGE texture wrap mode.

### DIFF
--- a/extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
+++ b/extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
@@ -25,8 +25,8 @@ Status
 
 Version
 
-    Last Modified Date: September 4, 2017
-    Revision 1
+    Last Modified Date: September 7, 2017
+    Revision 2
 
 Number
 
@@ -34,7 +34,7 @@ Number
 
 Dependencies
 
-    OpenGL ES 3.0 is required.
+    OpenGL ES 2.0 is required.
 
     This extension is written against the OpenGL ES 3.2 Specification.
 
@@ -103,6 +103,17 @@ Errors
     TEXTURE_WRAP_T, or TEXTURE_WRAP_R parameter is set to REPEAT,
     MIRRORED_REPEAT, or MIRROR_CLAMP_TO_EDGE.
 
+Dependencies on EXT_texture3D or equivalent
+
+    If EXT_texture3D or equivalent functionality is not implemented,
+    then the references to clamping of 3D textures in this file are
+    invalid, and references to TEXTURE_WRAP_R should be ignored.
+
+Dependencies on OpenGL ES 3.0 or equivalent
+
+    If OpenGL ES 3.0 or equivalent is not supported, then ignore all
+    references to sampler objects and SamplerParameter* functions.
+
 New State
 
     Only the type information changes for these parameters:
@@ -138,7 +149,10 @@ Issues
 
 Revision History
 
-    Revision 1, September 5, 2017 (criccio)
+    Revision 2 - September 7, 2017 (criccio)
+    - Require OpenGL ES 2.0 instead of OpenGL ES 3.0
+
+    Revision 1 - September 5, 2017 (criccio)
     - Initial EXT version based on ARB_texture_mirror_clamp_to_edge
 
 
--- a/extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
+++ b/extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
@@ -138,14 +138,7 @@ New Implementation Dependent State
 
 Issues
 
-    1. Should we include MIRROR_CLAMP for compatibility profiles?
-    What about MIRROR_CLAMP_TO_BORDER? (And more importantly can all
-    vendors support it?)
-
-    RESOLVED. No. Some vendors stated that they can't support
-    MIRROR_CLAMP_TO_BORDER and MIRROR_CLAMP (although it's not clear if
-    they'd ever ship a compatibility profile with this extension), but
-    regardless we'll leave them out of this extension.
+    None
 
 Revision History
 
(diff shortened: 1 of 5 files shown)


TobiasHector:
Is anyone actually exposing this, or planning on doing so? I don't see it on http://opengles.gpuinfo.org/gles_extensions.php. As a general rule of thumb, I don't think we should take extensions into the ES registry without at least one implementer signing up to do it. I mean we can, I'm just not sure what the point would be other than to confuse the ecosystem by implying something is supported when it's not?

I guess I'll put this on the next WG telecon agenda, and we'll see if anyone wants to sign up/comment.

Groovounet:
I am actually just hoping to get vendor onboard with this. Considering that https://vulkan.gpuinfo.org/listreports.php?extension=VK_KHR_sampler_mirror_clamp_to_edge is largely supported, I have good hopes.

ianromanick:
I'd be in favor of removing the ES 3.0 requirement.  I believe we have some GPUs that support GL_ARB_mirror_clamp_to_edge that only do ES 2.0.  That means an interaction section would need to be added that says, "If OpenGL ES 3.0 or GL_OES_texture3D is not supported, remove references to TEXTURE_WRAP_R."

Maybe this could replace the copy-and-paste issue about compatibility profile. :)

Groovounet:
Hi Ian,

I updated the extension proposal to require OpenGL ES 2.0 instead of OpenGL ES 3.0 and removed the copy pasted issue.

Thanks,
Christophe

extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
```diff
@@ -0,0 +1,151 @@
...
+Number
+
+    ES Extension #291
+
+Dependencies
+
+    OpenGL ES 2.0 is required.
+
+    This extension is written against the OpenGL ES 3.2 Specification.
+
```
dgkoch: This extension interacts with OES_texture_3D. 

This extension interacts with OpenGL ES 3.0.

This extension interacts with OpenGL ES 3.2.

extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
```diff
@@ -0,0 +1,151 @@
...
+    None
+
+Errors
+
+    The error INVALID_ENUM is generated when TexParameter* is called
+    with a target of TEXTURE_RECTANGLE and the TEXTURE_WRAP_S,
+    TEXTURE_WRAP_T, or TEXTURE_WRAP_R parameter is set to REPEAT,
+    MIRRORED_REPEAT, or MIRROR_CLAMP_TO_EDGE.
+
+Dependencies on EXT_texture3D or equivalent
```
dgkoch: OES_texture_3D (EXT_texture3D is an GL extension, not ES).

extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
```diff
@@ -0,0 +1,151 @@
...
+GLX Protocol
+
+    None
+
+Errors
+
+    The error INVALID_ENUM is generated when TexParameter* is called
+    with a target of TEXTURE_RECTANGLE and the TEXTURE_WRAP_S,
+    TEXTURE_WRAP_T, or TEXTURE_WRAP_R parameter is set to REPEAT,
+    MIRRORED_REPEAT, or MIRROR_CLAMP_TO_EDGE.
```
dgkoch: This error is not applicable for ES (no TEXTURE_RECTANGLE target).

extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
```diff
@@ -0,0 +1,151 @@
...
+
+    If EXT_texture3D or equivalent functionality is not implemented,
+    then the references to clamping of 3D textures in this file are
+    invalid, and references to TEXTURE_WRAP_R should be ignored.
+
+Dependencies on OpenGL ES 3.0 or equivalent
+
+    If OpenGL ES 3.0 or equivalent is not supported, then ignore all
+    references to sampler objects and SamplerParameter* functions.
+
```
dgkoch: Add: 
Dependencies on OpenGL ES 3.2 or equivalent

If OpenGL ES 3.2 or equivalent is not supported, then ignore all references to the TexParameterI* and SamplerParameterI* functions.

extensions/Makefile
```diff
@@ -17,7 +17,7 @@
 
 # python3 is required. Sorry, python2 dudes.
 
-PYTHON = python3
+PYTHON = python
```
dgkoch: I don't think this change should be included.

extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
```diff
@@ -0,0 +1,157 @@
...
+    Daniel Koch
+
+Notice
+
+    Copyright (c) 2017 The Khronos Group Inc. Copyright terms at
+        http://www.khronos.org/registry/speccopyright.html
+
+Status
+
+    Draft on September 7, 2017.
```
dgkoch: Did you mean to switch this to Draft?

dgkoch:
Looks good otherwise (and including the header changes).

Groovounet:
Erm, when I do "make" in the "extensions" directory it tells me "nothing to be done for 'all'". So the PHP files are not generated. Am I missing something?

extensions/EXT/EXT_texture_mirror_clamp_to_edge.txt
Groovounet: I mean it but fixed. 
Probably I should have post it for review as draft :D

oddhack:
@Groovounet it looks to me like the 'extensions' directory is up to date, which is why nothing happens - the generated PHP file is already current with registry.py, and part of the PR diff. Maybe you did it earlier and forgot?

dgkoch:
LGTM now. @oddhack please go ahead and merge this now.

I think the headers will still need to be re-generated after this is accepted

PR #100, opened 2017-09-06
Update GLX protocol through OpenGL 3.1

nwnk:
The old registry subversion repo had documentation for GLX protocol through OpenGL 3.1 in draft form. The drafts had rough consensus (see Khronos bzs 9885 and 14870) but never got properly merged, so let's fix that.

The new registry repo doesn't include the source for the GLX protocol spec document, so for the moment I've simply appended the new core request encoding for 2.x and 3.0 to glxproto.reserved.txt. For the remaining extension specs I've copied the GLX protocol from the old drafts. Xorg doesn't implement any of this yet. NVIDIA's GLX is documented as supporting at least some of these extensions in indirect contexts, so a double-check that these updates match the implementation would be most welcome.


oddhack:
It looks fine to me. Really if NVIDIA and X.org/Mesa agree we should just take it, but a formal vote of the WG would be nice. @pdaniell-nv can you call for that?

The GLX protocol spec can be regenerated at some point afterwards, certainly.

xml/gl.xml
```diff
@@ -9140,6 +9141,7 @@ typedef unsigned int GLhandleARB;
             <param><ptype>GLenum</ptype> <name>target</name></param>
             <param><ptype>GLuint</ptype> <name>index</name></param>
             <param><ptype>GLuint</ptype> <name>buffer</name></param>
+            <glx type="render" opcode="356"/>
```
cubanismo: Something seems to have gone wrong with the XML portion of the change BeginTransformFeedback has the right opcode, but at least a handful of the other ones I spot-checked don't from my reading of our internal driver code generation data and the specs I dug up on the old internal Khronos bugzilla.  Can you double-check these?

cubanismo:
@nwnk The changes to the extension specs look right, but something's wrong with the XML data.  See comment inline.

Our driver supports these, but since they're non-standard, it only enables them when the user has set some config options on both sides.  If this PR gets merged, I'll flip the switch to mark them standard in our next driver release.  If you want to test Mesa development against NVIDIA drivers in the meantime, set the env var __GL_ALLOW_UNOFFICIAL_PROTOCOL=1 to get our GLX client lib to send the protocol, and

  Option "AllowUnofficialGLXProtocol" "True"

In xorg.conf in your screen section to enable support in our GLX server lib.

xml/gl.xml
```diff
@@ -9177,6 +9179,7 @@ typedef unsigned int GLhandleARB;
             <param><ptype>GLuint</ptype> <name>buffer</name></param>
             <param group="BufferOffset"><ptype>GLintptr</ptype> <name>offset</name></param>
             <param group="BufferSize"><ptype>GLsizeiptr</ptype> <name>size</name></param>
+            <glx type="render" opcode="355"/>
```
cubanismo: As an example of a bad opcode here, this appears to assign render opcode 355 to glBindBufferBaseEXT, but that opcode is associated with glBindBufferRange() in our driver, and also in OpenGL3_0.txt in the old subversion tree and in a proposed update to NV_transform_feedback.txt I still have pending in NVIDIA's internal registry mirror, but can't find the Khronos bug for anymore.

xml/gl.xml
```diff
@@ -10034,24 +10037,28 @@ typedef unsigned int GLhandleARB;
             <param group="DrawBufferName"><ptype>GLint</ptype> <name>drawbuffer</name></param>
             <param><ptype>GLfloat</ptype> <name>depth</name></param>
             <param><ptype>GLint</ptype> <name>stencil</name></param>
+            <glx type="render" opcode="360"/>
```
cubanismo: In OpenGL3_0.txt, opcode 360 was assigned to glClearBufferfi, which I don't see included here.

nwnk:
```xml
<command>
    <proto>void <name>glClearBufferfi</name></proto>
    <param group="Buffer"><ptype>GLenum</ptype> <name>buffer</name></param>
    <param group="DrawBufferName"><ptype>GLint</ptype> <name>drawbuffer</name></param>
    <param><ptype>GLfloat</ptype> <name>depth</name></param>
    <param><ptype>GLint</ptype> <name>stencil</name></param>
    <glx type="render" opcode="360"/>
</command>
```

xml/gl.xml
cubanismo: Yeah, I think the github UI is lying to me.  When I click the little expand button, it makes it look like this is affecting BindBufferBase, but I see the line numbers jump about 30 lines and expand only gave me an additional line or two of context.  When I view the raw file it looks right.  Let me download and review properly in a terminal.

xml/gl.xml
cubanismo: OK, confirmed after some offline back-and-forth with @nwnk that this is me reading github webdiffs wrong.  After fetching & looking at the diff locally with more context, the opcodes look right.

cubanismo:
Update to my note above on testing against the NVIDIA driver: The opcodes we use for our "unofficial" protocol don't line up for the commands from NV_primitive_restart and ARB_uniform_buffer_object, but the rest should match the documentation AFAICT.

oddhack:
There seems to be a conflict to be resolved. Otherwise, @pdaniell-nv didn't say he was calling for a vote, and Mesa / NVIDIA are both represented here, so I'm comfortable merging it once the conflict's done with.

pdaniell-nv:
Yep, this is fine to merge.

Issue #94, opened 2017-08-09
GLX_MESA_swap_control missing in glx.xml


Dav1dde:
`GLX_EXT_swap_control` and `GLX_SGI_swap_control` are included in the specification, but [`GLX_MESA_swap_control`](https://github.com/mesa3d/mesa/blob/master/docs/specs/MESA_swap_control.spec) seems to be missing.

oddhack:
There is no registered specification for GLX_MESA_swap_control. If the Mesa folks want to provide one we can register it and add to the XML/headers, but without that we're out of luck. There are a number of other extant extensions that have no registered specs as well - I couldn't tell you the exact number because we only find out about them by accident.

I suggest you file this with the Mesa project, and ask them to open a PR here if they want to register it.

It's a reasonable issue but we have no way of resolving it, so I request you close this at least until there's something actionable associated with it.

oddhack:
The spec author pointed me to the master copy in the Mesa tree, so registered & closed now.

Issue #103, opened 2017-09-24
Macro GL_BLEND_COLOR not "required"


ferreiradaselva:
This is related with the issue https://github.com/Dav1dde/glad/issues/115.

I'm using [glad](https://github.com/Dav1dde/glad) loader, with OpenGL core 3.3, and it seems that the loader doesn't define `GL_BLEND_COLOR` because in the `gl.xml`, `GL_BLEND_COLOR` is not `<require>`'d by GL 2.0.

However, the functions `glBlendColor`, `glBlendFunc`, `glBlendEquation` are available for OpenGL 2.0-4.5, and assume that you can use `GL_BLEND_COLOR`.

Am I mistaking something or is it an actual bug?

oddhack:
I think that's a good catch, though I'm not actually going to go all the way back to 15 year old emails to verify what our reasoning was at the time, which I TBH don't remember :-) Added BLEND_EQUATION and BLEND_COLOR to GL 1.4 core feature interfaces in gl.xml, in #104 - let me know if that helps.

oddhack:
Closed by #104.

PR #104, opened 2017-09-24
Proposed fix for #103.

oddhack:
Apparently the state tokens BLEND_COLOR and BLEND_EQUATION weren't
included in the GL 1.4 core interfaces, although the corresponding entry
points were promoted from ARB_imaging to core. Also rearranged the subset of ARB_imaging tokens in both part of the XML to be in the same order, for slight clarity improvement.


oddhack:
Going to go ahead and merge this based on Piers' feedback. Also, I accidentally included a commit that fixes #94, and rather than tediously back that out into a separate PR I'm just going to take that as well.

PR #105, opened 2017-09-25
Add a new MESA_tile_raster_order extension.

anholt:
Some wording is new based on review feedback on the Mesa mailing list
-- version 1 had been looked at by Intel and AMD developers, and the
conclusion was that we needed to clarify TextureBarrier()'s role in
the list of exceptions given the new exception.

Implementation status: Kernel side ABI in place for vc4, piglit test
written and submitted, and Mesa implementation submitted.

### DIFF
--- a/api/GL/glext.h
+++ b/api/GL/glext.h
@@ -9219,6 +9219,13 @@ GLAPI void APIENTRY glResizeBuffersMESA (void);
 #define GL_MESA_shader_integer_functions 1
 #endif /* GL_MESA_shader_integer_functions */
 
+#ifndef GL_MESA_tile_raster_order
+#define GL_MESA_tile_raster_order 1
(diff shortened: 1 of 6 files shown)


anholt:
I've added the ES extension number now.  Do you also want a

    This extension is written against the GL_ARB_texture_barrier
    specification.

or something else?  As far as I'm concerned, it's ready to merge -- it's had all the Mesa feedback I would expect it to get.

oddhack:
@anholt we generally says something like "This extension is written against (what texture_barrier was written against), as modified by the GL_ARB_texture_barrier extension", or, alternatively put similar language in the individual "Modified" sections where appropriate. That would be great.

anholt:
The Mesa side is now reviewed and ready to merge.  Hopefully the new Dependencies text is good on this PR.

PR #108, opened 2017-10-02
add more bitfield group annotations and missing group to gl.xml

3b:


3b:
`InternalFormat` group contains enums `GL_RGB2`, `GL_R32_EXT`, and `GL_R32` which are not otherwise defined, should they be removed? Couldn't find any definitions outside of `gl.xml` either.

oddhack:
The R32 enums look like ancient cut & paste errors; I think it's safe to remove them.

oddhack:
This seems fine. Can you verify that there are no changes in the generated header files? That's really the only test we have.

3b:
Is `GL_RGB2` included in the "safe to remove" group?

There is one change that should affect the generated header files, I changed the `param` argument of `QueryObjectParameteruiAMD` to `GLbitfield` instead of `GLuint`. The [extension](https://www.khronos.org/registry/OpenGL/extensions/AMD/AMD_occlusion_query_event.txt) lists it as `uint`. Let me know if i should remove that part.

I'll see if i can build headers and check that later today.

oddhack:
We consider the vendor extension specifications to be authoritative (not to say they might not be wrong, just that we won't change their API without checking with them first, and reflecting that change in the extension spec). So best to leave that alone.

oddhack:
RGB2_EXT was in the original EXT_texture extension, which actually predates my time working at SGI and a subset of which went into GL 1.1. I think it's also safe to remove the non-EXT RGB2 token since that apparently wasn't in the subset.

N.b. it's entirely possible people will complain about any or all of these tokens being missing even though they're not part of any GL family API, in which case they'll have to go back in.

3b:
ok, with latest changes it doesn't change headers that i can see (only non-whitespace changes are generation date, and i think whitespace changes are just from line endings).

If the undefined enums have to go back in at some point, that is no problem. Just removed them to get rid of noise when checking for mismatches.

oddhack:
Er. OK, I didn't check this adequately before going in. There are a bunch of problems with the new groups, because you're redefining some enum values in the XML, which isn't allowed. The headers aren't modified but there are a lot of complaints in the build process. So, what needs to happen is to move the new \<enums> groups, such as PathFontStyle, into \<group> tags that just list the enums in the group without attempting to define their values.

PR #115, opened 2017-10-12
Add GL_QCOM_texture_foveated

tate-hornbeck:
Add QCOM_texture_foveated which allows foveation to be specified on textures

### DIFF
--- a/api/GL/glext.h
+++ b/api/GL/glext.h
@@ -51,7 +51,7 @@ extern "C" {
 #define GLAPI extern
 #endif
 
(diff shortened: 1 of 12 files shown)


extensions/arbext.php
```diff
@@ -1,422 +0,0 @@
-<ol>
```
oddhack: Did you do a 'make' in extensions/ to regenerate these files? Something appears to have messed up as AFAICT from the diff viewer, the .php files are empty.

tate-hornbeck:
Fixed the php files

PR #119, opened 2017-10-20
Fix #117 (missing needed DSA functionality in NV_path_rendering XML for GLES)

### LINKED ISSUE #117: gl.xml: GL_NV_path_rendering needs functionality from EXT_direct_state_access, not available in GLES
The extension for GL_NV_path_rendering states:

When NV_path_rendering is advertised, the following functionality must be supported...
The following entry points (specified by the EXT_direct_state_access extension) MUST be supported:

   void MatrixLoadfEXT(enum matrixMode, const float *m);
   void MatrixLoaddEXT(enum matrixMode, const double *m);
(snip)

However these entry points are never mentioned in gl.xml, and EXT_direct_state_access is not mentioned as a GLES extension. Maybe these entry points be added to gl.xml under this extension, instead of under EXT_direct_state_access, since that is not specified as a GLES extension.

This is making life hard for my auto-generation script. @oddhack any possible resolution?

oddhack:
Fixes #117 .

Unfortunately, a literal reading of the NV_path_rendering spec requires adding the double-precision DSA matrix commands to OpenGL ES, which I'm pretty sure is incorrect since OpenGL ES doesn't require support for double precision otherwise. We need to get clarification on this from NVIDIA, so I'll assign this to one of their Khronos people in the hope they will either respond, or pass on to someone who can.


pdaniell-nv:
It's not intended that GL_NV_path_rendering add 64-bit vertex attribute support to OpenGL ES.

@dgkoch I believe you added OpenGL ES support to this extension. Is there something we need to clarify here?

dgkoch:
> It's not intended that GL_NV_path_rendering add 64-bit vertex attribute support to OpenGL ES.

No, but it is intentional to add the double matrix functions in ES to support NV_path_rendering. We are aware that doubles aren't otherwise supported in ES, but these are just CPU-side operations.  If we ever do a multi-vendor version of the extension, we'd likely add proper float-only matrix entry points, but this is the way it works in our driver today.

dgkoch:
assigning @oddhack to rebase and merge.

oddhack:
Huh, OK. This is worthy of adding an Issue to the path_rendering spec, which I've done. Merging now.

PR #123, opened 2017-11-03
Noise functions are not present when generating SPIR-V.

johnkslang:
The noise*() functions in GLSL are deprecated and defined to return 0. But for SPIR-V, they should not be supported at all.


pdaniell-nv:
@oddhack We approved this change to merge.

PR #124, opened 2017-11-08
EXT_texture_sRGB_decode: add interaction with ARB_bindless_texture

nhaehnle:
Per the discussion on API issue 51. Attempting to put the rules into
words became a mess, so I replaced it all with a table.

Note I also added the 'statically' which I believe came up in
earlier discussions on texelFetch, to make it clear that the
implementation is not required to solve the halting problem.

### DIFF
--- a/extensions/EXT/EXT_texture_sRGB_decode.txt
+++ b/extensions/EXT/EXT_texture_sRGB_decode.txt
@@ -224,27 +224,27 @@ Additions to Chapter 3 of the 2.1 Specification (Rasterization)
 
    "The conversion of sRGB color space components to linear color space is
     always applied if the TEXTURE_SRGB_DECODE_EXT parameter is DECODE_EXT.
-    Table X.1 describes whether the conversion is applied if the
+    Table X.1 describes whether the conversion is skipped if the
     TEXTURE_SRGB_DECODE_EXT parameter is SKIP_DECODE_EXT, depending on
     the function used for the access, whether the access occurs through a
     bindless sampler, and whether the texture is statically accessed
     elsewhere with a texelFetch function."
 
     Add a new table X.1, Whether the conversion of sRGB color space
-    components to linear color space is applied when the
+    components to linear color space is skipped when the
     TEXTURE_SRGB_DECODE_EXT parameter is SKIP_DECODE_EXT.
 
                                        texelFetch       other builtin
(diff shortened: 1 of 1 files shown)


extensions/EXT/EXT_texture_sRGB_decode.txt
```diff
@@ -221,26 +223,33 @@ Additions to Chapter 3 of the 2.1 Specification (Rasterization)
...
+    bindless sampler, and whether the texture is statically accessed
+    elsewhere with a texelFetch function."
 
-    Otherwise, if the texel lookup function is one of the texture builtin
-    functions or one of the texture gather functions, the conversion of sRGB
-    color space components to linear color space is controlled by the
-    TEXTURE_SRGB_DECODE_EXT parameter.
+    Add a new table X.1, Whether the conversion of sRGB color space
+    components to linear color space is applied when the
+    TEXTURE_SRGB_DECODE_EXT parameter is SKIP_DECODE_EXT.
```
pdaniell-nv: Would the table be easier to understand if we reversed the sense. In other words the table is specified in terms of "DECODE_EXT" and not "SKIP_DECODE_EXT". At first look it's quite surprising.

extensions/EXT/EXT_texture_sRGB_decode.txt
```diff
@@ -221,26 +223,33 @@ Additions to Chapter 3 of the 2.1 Specification (Rasterization)
...
+    bindless sampler, and whether the texture is statically accessed
+    elsewhere with a texelFetch function."
 
-    Otherwise, if the texel lookup function is one of the texture builtin
-    functions or one of the texture gather functions, the conversion of sRGB
-    color space components to linear color space is controlled by the
-    TEXTURE_SRGB_DECODE_EXT parameter.
+    Add a new table X.1, Whether the conversion of sRGB color space
+    components to linear color space is skipped when the
+    TEXTURE_SRGB_DECODE_EXT parameter is SKIP_DECODE_EXT.
```
jackohound: When initially looking at this, I found the new table a little unclear.  After staring at this for a bit, i'm ok with it as-is.

gnl21: +1 I've had a look at this as well and the changes seem good to me.

PR #128, opened 2017-11-15
reserve MapBufferUsageMask 0x2000 for NVX_cross_process_interop

dgkoch:
Reserve buffer usage bit 0x2000 for NVX_cross_process_interop (which is not being published now). 
Also note that buffer usage bit 0x1000 was also previously reserved for NVIDIA.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -3474,6 +3474,8 @@ typedef unsigned int GLhandleARB;
         <enum value="0x0400" name="GL_SPARSE_STORAGE_BIT_ARB"/>
         <enum value="0x0800" name="GL_LGPU_SEPARATE_STORAGE_BIT_NVX"/>
         <enum value="0x0800" name="GL_PER_GPU_STORAGE_BIT_NV"/>
+            <unused start="0x1000" end="0x1000" comment="Reserved for NVIDIA"/>
+        <enum value="0x2000" name="GL_EXTERNAL_STORAGE_BIT_NVX"/>
     </enums>
 
     <enums namespace="GL" group="MemoryBarrierMask" type="bitmask">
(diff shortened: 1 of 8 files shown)


xml/gl.xml
```diff
@@ -3474,6 +3474,8 @@ typedef unsigned int GLhandleARB;
         <enum value="0x0400" name="GL_SPARSE_STORAGE_BIT_ARB"/>
         <enum value="0x0800" name="GL_LGPU_SEPARATE_STORAGE_BIT_NVX"/>
         <enum value="0x0800" name="GL_PER_GPU_STORAGE_BIT_NV"/>
+        <enum value="0x1000" name="reserved_0x1000_NV"/>
```
oddhack: Use the \<unused start="0x1000" end="0x1000" name="Reserved for NVIDIA"> syntax here until there's an actual enum defined, please.

xml/gl.xml
```diff
@@ -45806,6 +45808,11 @@ typedef unsigned int GLhandleARB;
             </require>
         </extension>
         <extension name="GL_NVX_blend_equation_advanced_multi_draw_buffers" supported="gl|gles2"/>
+        <extension name="GL_NVX_cross_process_interop" supported="gl|gles2">
+            <require comment="unpublished experimental extension">
```
oddhack: But the extension is being published now, at least in the sense that it will exist in the headers, right?

xml/gl.xml
dgkoch: will do, I didn't know about that.

xml/gl.xml
dgkoch: not really. there is no spec being published, and it's definitely not the whole extension interface. 
Ideally it would be nice to keep this out of the header, but still keep the token associated with the extension? The only reason I'm adding it is so the CTS can pick up the definition of the token.

oddhack: If you put supported="disabled" (or any regexp that doesn't match an actual api tag like "gl" or "gles2", then the extension will not be included in output headers. Same mechanism as Vulkan. (Though you should check to make sure it works as intended here, since the scripts have diverged quite a lot).

xml/gl.xml
dgkoch: done.

xml/gl.xml
dgkoch: "disabled" does seem to work, and the CTS scripts are still happy with it as well.

dgkoch:
@oddhack this is ready for re-review now. I rebased it as well, and it looks like there were some pending header changes that this picked up now.

PR #131, opened 2017-11-21
WIP: Add KHR_EGL_image_storage extension.

tweenk:
This extension allows the creation of immutable texture EGLImage targets. These have much simpler semantics than regular texture EGLImage targets defined by OES_EGL_image and do not exhibit potentially surprising orphaning behavior.

We need this to more easily support new AHardwareBuffer functionality in Android P. Since this extension does contain anything Android-specific and looks like it could be useful to other users of EGLImage, we would like to ratify it as a KHR extension.

### DIFF
--- a/api/GL/glcorearb.h
+++ b/api/GL/glcorearb.h
@@ -3858,6 +3858,17 @@ GLAPI void APIENTRY glTexBufferARB (GLenum target, GLenum internalformat, GLuint
 #define GL_ARB_viewport_array 1
 #endif /* GL_ARB_viewport_array */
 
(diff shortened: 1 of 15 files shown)


oddhack:
@tweenk please bring this up inside Khronos on the GL and/or GL ES mailing lists - that's the only way to get a KHR extension ratified.

oddhack:
Marking WIP so this doesn't get merged until the GL working groups can sign off and get it ratified.

pdaniell-nv:
Closing to handle this internally.

PR #149, opened 2018-02-02
Include the API name in the header guard.

c0d1f1ed:
Several headers have the same header guard, making it impossible to
include declarations from both headers. This was complicating the
implementation of a driver with support for both desktop OpenGL and
OpenGL ES.

### DIFF
--- a/xml/reg.py
+++ b/xml/reg.py
@@ -513,7 +513,7 @@ class COutputGenerator(OutputGenerator):
         #
         # Multiple inclusion protection & C++ wrappers.
         if (genOpts.protectFile and self.genOpts.filename):
-            headerSym = '__' + re.sub('\.h', '_h_', os.path.basename(self.genOpts.filename))
+            headerSym = '__' + self.genOpts.apiname + '_' + re.sub('\.h', '_h_', os.path.basename(self.genOpts.filename))
             write('#ifndef', headerSym, file=self.outFile)
             write('#define', headerSym, '1', file=self.outFile)
             self.newline()


pdaniell-nv:
I want to run this change by the working group to make sure there are no objections. It seems like the right thing to do.

api/GLES3/gl3.h
```diff
@@ -1,5 +1,5 @@
-#ifndef __gl3_h_
-#define __gl3_h_ 1
+#ifndef __gles2_gl3_h_
```
gnl21: Should this say "gles3_gl3_h"?

c0d1f1ed: Changing the 'apiname' to 'gles3' in genheaders.py also causes changes like no longer defining GL_ES_VERSION_2_0 in gl3.h. So that's not going to work. Note that despite the apparent inconsistency, this header guard does not clash with any others so it doesn't need fixing.

I think the way we need to interpret it is that 'apiname' and API version are completely different things. They might as well have called it 'glesa' and 'glesb' and each of them support different version numbers. Note also that OpenGL ES 3.x is recommended to be implemented in a libGLESv2 library.

c0d1f1ed:
Thanks for the feedback. I don't think we should be concerned about those things. Using header guard symbols for any other purpose would be extremely dubious coding practice that can be expected to break, and thus doesn't justify holding back fixes. The only valid usage I can think of would be to undef it so that another header with the same guard can be included, i.e. to fix this very issue, and importantly that pattern would not be broken by this change.

Including both ES and GL headers in the same app is valid, or at least should be, because some extensions that were originally defined for GL have also been implemented by ES drivers. Also, some apps don't know if they'll have a functional GL and/or ES driver available, and the selection can be based on a frequently updated blacklist/whitelist (e.g. Chrome does this).

Note that the workaround of undefining the header guard is in fact very fragile, so I don't consider that viable. Generating a header with both sets of interfaces seems like a maintenance burden, and could also break for numerous reasons. In my opinion only when the apps and the driver implementations include the same headers do we have reasonable guarantees of compatibility.

pdaniell-nv:
@oddhack Touched on again this week. The WG approves this PR.

c0d1f1ed:
Thanks Piers. Anything else we're waiting on here to be able to merge this?

pdaniell-nv:
No, nothing else you need to do. @oddhack will merge at some point.

c0d1f1ed:
Rebased.

1ace:
Related to this issue, we just had a [bug report in Mesa](https://bugs.freedesktop.org/show_bug.cgi?id=105328) about conflicting declarations in GL & GLES headers.
The gist of it is:
```
/usr/include/GLES3/gl31.h:77:25: error: conflicting declaration 'typedef khronos_ssize_t GLsizeiptr'
 typedef khronos_ssize_t GLsizeiptr;
                         ^~~~~~~~~~
/usr/include/GL/glext.h:468:19: note: previous declaration as 'typedef ptrdiff_t GLsizeiptr'
 typedef ptrdiff_t GLsizeiptr;
                   ^~~~~~~~~~
```

Fixing the header guards in this PR might cause more conflicts like this.

I'll raise a separate issue for this specific conflict, but I just wanted to mention it here so that it can be verified before merging.

c0d1f1ed:
Thanks for raising that concern. I've checked that it's not an issue. The only headers currently sharing the same guard are GL/glext.h and GLES/gl.h, and my change allows them to both be included without conflicts.

Also note that even if conflicts were caused, it's easy to work around them by not including the second header, because that's effectively what happens today when including a header which uses the same guard. So I think #162 is orthogonal to this.

@oddhack is anything still keeping us from merging this? It's obviously a bug to have headers with the same include guard.

oddhack:
@c0d1f1ed could you revert the header file changes and just provide the script change? I'll accept it and regenerate them myself. Unfortunately the generated files are prone to conflicts whenever PRs get queued up and dealt with in a batch, as is happening tonight :-(

c0d1f1ed:
@oddhack Done. Thanks!

1ace:
@oddhack I think you forgot to regen the headers when you merged this PR?
Or did you mean that there's a script that will do this automatically tonight?

oddhack:
I forgot to push the changes. Done now.

PR #150, opened 2018-02-05
Add GL_AMD_gpu_shader_half_float_fetch extension spec

DominikWitczakAMD:


oscarbg:
time to publish also  VK_AMD_gpu_shader_half_float_fetch? this is shipping in Vega drivers and not public yet? also what's the status on glslang AMD_gpu_shader_half_float_fetch support?

oddhack:
Vendor extensions to Vulkan get published when (a) they've been approved internally for merge and (b) we do a public spec repository update, so even if such an extension were ready, it would take longer to appear.

PR #155, opened 2018-02-14
Fix HSL ClipColor() equation in the blend_equation_advanced extensions

pdaniell-nv:
Fixes https://www.khronos.org/members/login/bugzilla/show_bug.cgi?id=16026 and https://gitlab.khronos.org/opengl/API/issues/60.

Fix the HSL blend equation ClipColor() pseudo-code and modify the spec to say that the HSL blend equations are only well defined when the input color components are in the range [0..1].

Once this is approved we need to update the OpenGL ES 3.2 spec with the same change.

### DIFF
--- a/extensions/KHR/KHR_blend_equation_advanced.txt
+++ b/extensions/KHR/KHR_blend_equation_advanced.txt
@@ -985,7 +985,7 @@ Revision History
 
       Fix ClipColor() equation where in the "if (maxcol > 1.0)" body the
       "(color-lum)*lum" term should have been "(color-lum)*(1-lum)". Also
-      add a new issue for the case where the inputs to SetLum() are outside
+      add new issue 11 for the case where the inputs to SetLum() are outside
       the range [0..1] and could cause a divide-by-zero in ClipColor().
 
     Revision 16, April 16, 2016 (from a September 30, 2014 edit that wasn't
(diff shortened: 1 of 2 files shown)


extensions/KHR/KHR_blend_equation_advanced.txt
```diff
@@ -959,8 +961,27 @@ Issues
       optimizations or include logic to check for zero alpha values for each
       input.
 
+    (35) How should ClipColor() behave if the inputs cause a divide-by-zero,
```
nvpbrown: I would suggest writing this as:

(35) For "HSL" blend equations, the blend equation involves a clipping step where colors may be "clipped" if the blend would produce components are outside the range [0,1].  Are there inputs where this blend could produce ill-defined or nonsensical results?

RESOLVED:  Yes, the results of HSL blend equations are undefined if the input colors have components outside the range [0,1].  Even if the input colors are in-range, the basic color adjustment done in these blends could produce result components outside the range [0,1].  To compensate, the ClipColor() function in the specification interpolates the result color and a greyscale value that matches the luminance of the result.  The math for the clipping operation assumes the luminance of the result color is in the range [0,1].   If that isn't the case, the clipping operation could result in a divide by zero (when all result components have the same out-of-bounds value) or perform an otherwise nonsensical computation.

pdaniell-nv:
Thanks for the suggestion. Updated extensions with the new issue language.

mnetsch:
We would like more time to confirm the new equation, holding off thumbs up for now.

mnetsch:
+1

pdaniell-nv:
@oddhack This is now approved to merge. Thanks.

PR #161, opened 2018-02-27
Rename parameters of glDepthRangef to "n" and "f"

nicebyte:
"near" and "far" conflict with macros defined by windows.h

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -13785,8 +13785,8 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glDepthRange</name></proto>
-            <param><ptype>GLdouble</ptype> <name>near</name></param>
-            <param><ptype>GLdouble</ptype> <name>far</name></param>
+            <param><ptype>GLdouble</ptype> <name>n</name></param>
+            <param><ptype>GLdouble</ptype> <name>f</name></param>
             <glx type="render" opcode="174"/>
         </command>
         <command>


pdaniell-nv:
There are other places where "n" and "f" are used. Don't you want to fix those ones too?

nicebyte:
I searched through gl.xml and this was the only place that used "near" and "far". Are there any other files I may not know about?

pdaniell-nv:
Doh, you're right. My brain thought you were doing n->near rather than near->n. Looks like the other functions like glDepthRangeIndexed already use "n" and "f".

oddhack:
Seems fine. I don't recall the history of this; the specs have used n/f forever but because the XML and specs are marked up separately (unlike e.g. Vulkan where there's only one source for prototypes), inconsistencies can happen. I am a bit surprised nobody on Windows ran into this in the last however-many-years it's been in the XML, though.

nicebyte:
Yeah, I'm surprised about it too. Though maybe it's because lots of people use GLEW or gl headers generated by other tools and the authors of said tools just patched the issue on their end.

pdaniell-nv:
OpenGL/ES WGs approve to merge.

oddhack:
I'll go ahead and accept this and regenerate the headers now. The next time we do a public spec update, it will show up there as well.

oddhack:
Woops, never mind. The API spec already uses 'n' / 'f'. We must have done the fix there at some point and forgotten to propagate it to XML/headers.

PR #166, opened 2018-03-07
Adding len value to gl.xml

B1m0:
Adding missing length information values for pointer parameters of the disjoint timer query and robustness extension functions


pdaniell-nv:
Assigned to @oddhack to review and merge. The WG is happy with the principle.

oddhack:
Seems fine. If anyone is actually using that tag they'll have to implement a command-specific COMPSIZE() function themselves, of course. SGI used to do that in its internal specfile system, many years ago.

PR #175, opened 2018-04-17
Fix link name for GLX_EXT_stereo_tree in registry

cubanismo:
For some reason GLX_EXT_stereo_tree was listed on
the registry page as GL_EXT_glx_stereo_tree.  This
change just corrects it to GLX_EXT_stereo_tree.


cubanismo:
Added missing file to the commit in the last update.

PR #178, opened 2018-05-09
Clarify arrays of arrays is allowed for resources

pdaniell-nv:
This resolves internal issue https://gitlab.khronos.org/opengl/API/issues/70.

This clarifies that OpenGL allows SPIR-V shaders to use multi-dimensional arrays for resources like UBOs and SSBOs, in that same way that OpenGL allows this for GLSL shaders. This is a difference to the SPIR-V environment for Vulkan, which only allows single-dimension arrays for these resources.


pdaniell-nv:
@johnkslang let me know if you're happy with this wording. I wanted to be precise about what is allowed and now allowed and where.

pdaniell-nv:
Assigned to @oddhack to merge.

PR #179, opened 2018-05-10
OVR_multiview version #4

jhelferty-nv:
Edits to OVR_multiview for desktop. Whenever a multiview framebuffer is bound, these edits make the results of glBegin/ glEnd undefined and the contents of gl_Layer in the fragment shader undefined. Also
clarifies what happens when a transform feedback object is paused.


pdaniell-nv:
This has been approved to merge internally https://gitlab.khronos.org/opengl/registry/merge_requests/13.

PR #180, opened 2018-05-10
GLX_OML_sync_control: Remove spurious GLXFBConfig attribute

nwnk:
This seems to just be an error. GLX_SBC is not in fact a defined token,
and it's not clear what its value would even be, as the SBC is a
property of the drawable not the fbconfig.


pdaniell-nv:
Assigned to @oddhack to merge.

PR #186, opened 2018-06-03
Remove redundant type definitions in gl.xml now that khrplatform.h is required for all headers

### LINKED ISSUE #184: GLsizeiptr/GLintptr type change, depending on khrplatform
Followup for: https://github.com/KhronosGroup/OpenGL-Registry/issues/162

1) Due to the change the GL API depends on `khrplatform.h` now which was previously not the case, this seems like a breaking change that may affect multiple users of the specification. This may also affect users that are using `glcorearb.h` directly (including OS package maintainers?).

2) The typedefinitions have a lot of duplicates now `<type requires="khrplatform">typedef khronos_intptr_t <name>GLintptr</name>;</type>`: `<type api="gles1" requires="khrplatform">typedef khronos_intptr_t <name>GLintptr</name>;</type>`, `<type api="gles2" requires="khrplatform">typedef khronos_intptr_t <name>GLintptr</name>;</type>` etc.

3) Extension types were not changed: `<type requires="stddef">typedef ptrdiff_t <name>GLintptrARB</name>;</type>`

Background information (shouldn't affect your decision on keeping this change):
[glad](https://github.com/Dav1dde/glad) an application to generate loaders based on these specifiactions broke, mainly due to a bug that was triggered by this and not the specification change itself. That's how I noticed the change (multiple users reported the generator breaking over night). My fix was to wrap the typedefs with a macro that checks if `khrplatform.h` was previously included, this works for glad but would not be enough for the specs.

I am not sure to fix this "better", it might be possible using macros, like guarding the type definition.

oddhack:
The first commit in this branch just does what was discussed in #184, using a common definition for GLintptr / GLsizeiptr. However, now that khrplatform.h is required everywhere, the final state of the branch uses the khronos_* sized types more broadly to define GL types. Arguably there are a few cases I left out, but making GLint be anything other than platform 'int' might be problematic on some platforms. OTOH GLhalf must be a 16-bit type. Etc.

Should get signoff from the GL WG and Mesa before merging.

Fixes #184


Dav1dde:
What's with the `inttypes` type, is it still necessary? Looks like `khrplatform.h` already has that logic.

Other than that, looks good to me.

oddhack:
@Dav1dde done. Thanks for catching that. I don't have any dead XML tag elimination tools.

api/GL/glext.h
```diff
@@ -464,7 +464,6 @@ GLAPI void APIENTRY glBlendEquation (GLenum mode);
 
 #ifndef GL_VERSION_1_5
 #define GL_VERSION_1_5 1
-#include <KHR/khrplatform.h>
```
c0d1f1ed: This include is still needed. Else the next couple of lines miss the definition for the khronos_* types.

pdaniell-nv:
OpenGL/ES WGs have no objection to this change in principle. Once everyone here is happy it's approved for merge.

c0d1f1ed:
@pdaniell-nv Please note my comment above. This doesn't compile on Windows:

#include <windows.h>
#include <GL/GL.h>
#include <GL/glext.h>

GL/glext.h(467): error C4430: missing type specifier

pdaniell-nv:
How do we unblock this issue? It's not clear to me what the next action should be. It seems like we need to resolve the Windows issue. Anything else?

oddhack:
@pdaniell-nv I see the problem - the dependency generation thinks the khrplatform.h include was satisfied by the GL 1.0 interfaces, but, those interfaces aren't emitted in glext.h since it only includes GL 1.2+ extensions. Fixing it properly will require some thought. I'm not sure we've encountered this situation previously. I can do a quick & dirty fix by putting the khrplatform.h include in the boilerplate following the comments at start of file, though, which maybe would suffice for the time being.

oddhack:
@pdaniell-nv @c0d1f1ed please try now, I think glext.h includes khrplatform.h properly.

c0d1f1ed:
That works for us, thanks!

oddhack:
Given @pdaniell-nv's previous approval, I am merging this.

api/GL/glext.h
oddhack: Fixed

PR #188, opened 2018-06-15
Wrong value for the WGL_EXT_colorspace extension

timceesq:
The value for WGL_COLORSPACE_EXT is 0x309D as per extension's specification, not 0x3087. Also, tested on the newest NVIDIA drivers and only the value 0x309D is correctly recognized.


oddhack:
Thanks. I see the problem - when the original XML entry was constructed someone (probably me) used the enum value for EGL_COLORSPACE instead of EGL_GL_COLORSPACE, following the comments on aliasing of EGL enums a bit too literally :-(

diaphore:
The problem persists : [master/xml/wgl.xml#L454](https://github.com/KhronosGroup/OpenGL-Registry/blob/master/xml/wgl.xml#L454), I'm guessing the modification to `wglext.h` was overwritten.

PR #190, opened 2018-06-24
gl.xml: Remove compatibility profile from GL_VERSION_3_0

Rijenkii:


pdaniell-nv:
The OpenGL/ES WG approves this change as long as @oddhack is happy.

oddhack:
N.b. github's diff viewer is broken with files this large, IMO - it shows lines from two very different line ranges as contiguous, which can be very confusing. I reported this a while ago but no action yet :-(

Rijenkii:
@oddhack Done!

PR #191, opened 2018-06-28
Alias glGetGraphicsResetStatusEXT to core function

tate-hornbeck:
Support contexts that only expose GL_EXT_robustness


tate-hornbeck:
This is to address a CTS issue https://gerrit.khronos.org/#/c/2673/ .

jackohound:
This change was requested in Khronos-internal [Gerrit 2673](https://gerrit.khronos.org/#/c/2673/).   CTS uses the XML file to autogenerate code.

pdaniell-nv:
OpenGL/ES WG approves this change.

PR #194, opened 2018-07-18
Add MESA_framebuffer_flip_y extension

fritzk:
Adds an extension to glFramebufferParameteri
that will specify if the framebuffer is vertically
flipped. Historically system framebuffers are
vertically flipped and user framebuffers are not.
Checking to see the state was done by looking at
the name field.  This adds an explicit field.


fritzk:
This extension has been moving through the mesa list.  (https://lists.freedesktop.org/archives/mesa-dev/2018-July/199946.html)

It is at the point where there isn't any objection to the idea and I would like to begin formalizing it.

chadversary:
+1

extensions/MESA/MESA_framebuffer_flip_y.txt
```diff
@@ -0,0 +1,81 @@
...
+
+    Accepted by the <pname> argument of FramebufferParameteri and
+    GetFramebufferParameteriv:
+
+        GL_FRAMEBUFFER_FLIP_Y_MESA                      0x8BBB
+
+Errors
+
+    An INVALID_OPERATION error is generated by GetFramebufferParameteriv if
+    the default framebuffer is bound to and is FRAMEBUFFER_FLIP_Y_MESA.
```
chadversary: Oops. Github's Markdown mangled my earlier reply. This block should read:

`An INVALID_OPERATION error is generated by GetFramebufferParameteriv if
the default framebuffer is bound to <target> and <pname> is FRAMEBUFFER_FLIP_Y_MESA.`

oddhack:
From a mechanics view, this is fine - just a question of when you want it merged. Is @chadversary the only person you want signing off on the spec aspects before doing that?

extensions/MESA/MESA_framebuffer_flip_y.txt
fritzk: thanks.  i thought it sounded odd, should have spoken up.

fritzk:
@chadversary is the only one needed for sign-ff

PR #197, opened 2018-07-25
Fix off-by-one in baseViewIndex + numViews check

Oletus:
If baseViewIndex + numViews is equal to the value of MAX_ARRAY_TEXTURE_LAYERS, then the layer pointed to by the last view can still fit into a 2D array texture.

### DIFF
--- a/extensions/OVR/OVR_multiview.txt
+++ b/extensions/OVR/OVR_multiview.txt
@@ -258,8 +258,7 @@ and Framebuffer Objects)
     An INVALID_VALUE error is generated if:
     - <numViews> is less than 1 or if <numViews> is greater than MAX_VIEWS_OVR.
     - <texture> is a two-dimensional array texture and <baseViewIndex> +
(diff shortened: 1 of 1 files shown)


extensions/OVR/OVR_multiview.txt
```diff
@@ -778,4 +777,5 @@ Revision History
                                 interop with compatibility profile's Begin/End (34). Add mention
                                 of OES_geometry_shader and OES_tessellation_shader to interop.
                                 Clarify what happens when transform feedback is paused.
+      5     07/25/18  oetuaho   Fix off-by-one issue in baseViewIndex + numViews check.
```
pdaniell-nv: Don't forget to change the revision and date at the top of the spec.

Oletus: Done.

pdaniell-nv:
We approved this change in the OpenGL/ES meeting today.

PR #207, opened 2018-08-20
Add QCOM_shader_framebuffer_fetch_rate

jackohound:
Adds extension QCOM_shader_framebuffer_fetch_rate, which was discussed in Khronos internal [WG-Docs#8](https://gitlab.khronos.org/opengl/WG-Docs/issues/8).    This PR also orders the QCOM extensions in alphabetical order in registry.py.

PR #210, opened 2018-08-27
Adding NV_memory_attachment

crohdewm:


oddhack:
Please revert the generated headers from the PR to minimize conflicts. Also, note that you're in a race condition with #207, which wants to use the same OpenGL ES extension number. I suggest incrementing your number by one, to prevent conflicts.

extensions/registry.py
```diff
@@ -3410,6 +3410,13 @@
         'supporters' : { 'NVIDIA' },
         'url' : 'extensions/NV/NV_light_max_exponent.txt',
     },
+    'GL_NV_memory_attachment' : {
+        'number' : 524,
+        'esnumber' : 304,
```
oddhack: Also needs to change to 305 to match the extension spec document. Otherwise this LGTM.

oddhack:
Unfortunately esext.php will also need to be reverted (or resolved) due to the other ES extension that just went in - then I can accept this.

oddhack:
Headers updated.

PR #218, opened 2018-09-27
Fix NVIDIA Turing extensions to appear in glcorearb.h.

nvpbrown:
All of the NVIDIA Turing extensions are supported in the core
profile.  But they currently don't appear in glcorearb.h because I
used 'supported="gl"' instead of 'supported="gl|glcore"' in gl.xml.

@oddhack @pixeljetstream


oddhack:
Header updated in 90be05229bf47a8803b1e242b860ad1dcfe1f3a0

PR #219, opened 2018-10-06
Missing enums

StailGot:
fix https://github.com/KhronosGroup/OpenGL-Registry/issues/220


NogginBops:
@StailGot sorry to revive this 4 years after the fact, but do you remember why you removed `GL_LOGIC_OP` from the `BlendEquationModeEXT` group? It's supposed to be in that group through the `EXT_blend_logic_op` extension from what I can see.

PR #222, opened 2018-10-12
Define tokens for VIEW_CLASS_{EAC*,ETC2*,ASTC*}

dgkoch:
Add view compatibility classes for EAC/ETC2 and ASTC formats
and add them for ARB_internalformat_query2.
These are mentioned in EXT/OES_texture_view for ES, but they
are never externally visibile and thus weren't assigned tokens.
However we also support these formats natively in GL on some
platforms, so we need tokens to return when the VIEW_COMPATIBILITY_CLASS
is queried on those platforms.

Reserve tokens for the VIEW_CLASSes for ASTC_3D, but don't
actually enable them since there are no extensions which
add these for GL.

### DIFF
--- a/extensions/ARB/ARB_internalformat_query2.txt
+++ b/extensions/ARB/ARB_internalformat_query2.txt
@@ -33,8 +33,8 @@ Status
 
 Version
 
-    Last Modified Date: February 14, 2018
-    Revision: 19
+    Last Modified Date: October 12, 2018
+    Revision: 20
 
 Number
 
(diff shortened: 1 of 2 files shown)


oddhack:
Is this ready to merge?

pdaniell-nv:
We'll touch on this in the next OpenGL/ES meeting and approve it there.

pdaniell-nv:
Discussed in the OpenGL/ES meeting today. Approved to merge.

oddhack:
Accepted. I'm adding a couple of comments to the XML to make clear which enums are only supported when corresponding extensions are supported, which has no effect on generated headers.

PR #232, opened 2018-12-07
Make use of 'work group' consistent across ARB/EXT specs

oddhack:
Fixes internal Bugzilla 11723 for ARB / EXT / OES specs (already fixed in core API specs).

N.b. there appear to be many remaining instances of inconsistent use of 'work group' in NV extension specs, which they can fix if they want to.

There are some incidental whitespace diffs.


dgkoch:
(the rest of the changes that you have here look fine though).

oddhack:
@pdaniell-nv I mercifully forgot about that. Surprised nobody has complained about WORK_GROUP vs. "workgroup" yet. I will update the PR.

oddhack:
@pdaniell-nv @dgkoch please re-review.

PR #236, opened 2019-01-10
Explicitly give rules for SPIR-V uniform offsets. 

oddhack:
These changes will be reflected in the next OpenGL 4.6 spec update as well, once that's published.


johnkslang:
This seems to drop change 40, which doesn't seem right.

oddhack:
Thanks, should be fixed now.

johnkslang:
The change log is fixed, but the functionality on line 361 is still not?

oddhack:
Should be OK now.

PR #240, opened 2019-01-24
Register EXT_depth_clamp for GLES

gerddie:
This extension provides the same interaction for OpenGL ES that ARB_depth_clamp does for OpenGL.

I went for an EXT extension because it is based on this already available ARB extension, and for that reason it makes sense to me to get it fully approved, but I could also put it in the MESA namespace (since there I will implement it). Many thanks for any comments.

### DIFF
--- a/extensions/EXT/EXT_depth_clamp.txt
+++ b/extensions/EXT/EXT_depth_clamp.txt
@@ -1,0 +1,148 @@
+Name
+
+    EXT_depth_clamp
+
+Name Strings
+
+    GL_EXT_depth_clamp
+
+Contact
+
+    Gert Wollny (gert.wollny 'at' collabora.com)
+
+Notice
+
+    Copyright (c) 2019 Collabora LTD 
+    Copyright (c) 2009-2013 The Khronos Group Inc. Copyright terms at
+        http://www.khronos.org/registry/speccopyright.html
+
+Status
+
+    FIXME
+
+Version
+
+    Version 1, 2019/01/24.
+    Based on ARB_depth_clamp version 4, modified 2009/08/02.
+
+Number
+
+    #309
+
+Dependencies
+
+    OpenGL ES 2.0 is required.
+    Written based on the wording of the OpenGL ES 3.2 specification.
+
+Overview
+
+    Conventional OpenGL clips geometric primitives to a clip volume
+    with six faces, two of which are the near and far clip planes.
+    Clipping to the near and far planes of the clip volume ensures that
+    interpolated depth values (after the depth range transform) must be
+    in the [0,1] range.
+
+    In some rendering applications such as shadow volumes, it is useful
+    to allow line and polygon primitives to be rasterized without
+    clipping the primitive to the near or far clip volume planes (side
+    clip volume planes clip normally).  Without the near and far clip
+    planes, rasterization (pixel coverage determination) in X and Y
+    can proceed normally if we ignore the near and far clip planes.
+    The one major issue is that fragments of a  primitive may extend
+    beyond the conventional window space depth range for depth values
+    (typically the range [0,1]).  Rather than discarding fragments that
+    defy the window space depth range (effectively what near and far
+    plane clipping accomplish), the depth values can be clamped to the
+    current depth range.
+
+    This extension provides exactly such functionality.  This
+    functionality is useful to obviate the need for near plane capping
+    of stenciled shadow volumes.  The functionality may also be useful
+    for rendering geometry "beyond" the far plane if an alternative
+    algorithm (rather than depth testing) for hidden surface removal is
+    applied to such geometry (specifically, the painter's algorithm).
+    Similar situations at the near clip plane can be avoided at the
+    near clip plane where apparently solid objects can be "seen through"
+    if they intersect the near clip plane.
+
+New Procedures and Functions
+
+    None
+
+New Tokens
+
+    Accepted by the <cap> parameter of Enable, Disable, and IsEnabled,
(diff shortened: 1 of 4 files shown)


xml/gl.xml
```diff
@@ -43292,6 +43292,11 @@ typedef unsigned int GLhandleARB;
                 <command name="glDepthBoundsEXT"/>
             </require>
         </extension>
+        <extension name="GL_EXT_depth_clamp" supported="gles2">
+            <require>
+                <enum name="GL_DEPTH_CLAMP"/>
```
kusma: I *think* this should be `GL_DEPTH_CLAMP_EXT`, no?

...Which would also require a line of this sort further up in this file (after the non-EXT one or so, I would think):
`<enum value="0x864F" name="GL_DEPTH_CLAMP_EXT"/>`

At least I don't think I've ever seen an EXT extension use a non-suffixed enum before... But then again, EXT extensions usually pop up before the functionality goes into the core, so perhaps that dance is pointless here...?

gerddie: I've changed it accordingly, thanks.

extensions/EXT/EXT_depth_clamp.txt
```diff
@@ -0,0 +1,222 @@
...
+    None
+
+New State
+
+Add to table 6.4, transformation state
+
+Get Value       Type  Get Command  Initial Value  Description     Sec      Attribute
+--------------  ----  -----------  -------------  --------------  ------   ----------------
+DEPTH_CLAMP     B     IsEnabled    False          Depth clamping  12.5   transform/enable
+                                                  on/off
```
dgkoch: DEPTH_CLAMP_EXT here as well. 

You should also remove the Attribute column as that is not applicable in ES.

extensions/EXT/EXT_depth_clamp.txt
```diff
@@ -0,0 +1,222 @@
...
+
+New Procedures and Functions
+
+    None
+
+New Tokens
+
+    Accepted by the <cap> parameter of Enable, Disable, and IsEnabled,
+    and by the <pname> parameter of GetBooleanv, GetIntegerv,
+    GetFloatv, and GetDoublev:
```
dgkoch: GetDoublev doesn't exist in ES.

extensions/EXT/EXT_depth_clamp.txt
```diff
@@ -0,0 +1,222 @@
...
+
+Additions to Chapter 12 of the OpenGL ES 3.2 Specification (Fixed-Function Vertex Post-Processing)
+
+ --  Section 12.5 "Primitive Clipping"
+
+    Add to the end of the 1st paragraph:
+
+    "Depth clamping is enabled with the generic Enable command and
+    disabled with the Disable command. The value of the argument to
+    either command is DEPTH_CLAMP. If depth clamping is enabled, the
```
dgkoch: DEPTH_CLAMP_EXT

extensions/EXT/EXT_depth_clamp.txt
```diff
@@ -0,0 +1,222 @@
...
+      However, if depth range is subsequently set to 0.2 and 0.8 and
+      depth clamping is enabled, the bitmap fragments will have their
+      zw depth component clamped to 0.8.  But if the depth range was
+      changed to 0.2 and 0.8 but depth range clamped is disabled, the
+      bitmap fragments will have a 0.9 zw depth component since then
+      the depth clamping is then not applied.
+
+    What push/pop attrib bits should affect the depth clamp enable?
+
+      RESOLUTION:  GL_ENABLE_BIT and GL_TRANSFORM_BIT.
```
dgkoch: This issue not applicable for ES. 

What I'd suggest instead of repeating all the issues from the earlier versions of the extension, is just say
"See the issue list in GL_ARB_depth_clamp" and only include issues unique to this extension. 

It's also a good idea to include an issue "How does this extension differ from ARB_depth_clamp", and describe any differences (of which there aren't much). 

In newer extensions, the issues list also goes at the end, just before the revision history.

extensions/EXT/EXT_depth_clamp.txt
gerddie: done

extensions/EXT/EXT_depth_clamp.txt
gerddie: changed accordingly.

gerddie:
Thanks for the review, I've force pushed a new version, but I didn't update the history, because I'm not sure whether this would be needed at this point. I hope that my changes are what you would expect, and I'm thankful for additional pointers if needed.

gerddie:
@oddhack done, the feedback should also be resolved, I'm not sure why github is still putting that marker there, the attribute column that the first comment mentions has been removed.

dgkoch:
Adding @pdaniell-nv to add to the ES WG agenda to see if there is any other feedback/interest from other ES vendors. 

@gerddie do you have any tests planned for this?

gerddie:
@dgkoch I'm not sure what tests you mean, but we want to implement this in mesa to be able to run OpenGL in a virtualised environment on top on a host that supports only GLES via [virglrenderer](https://gitlab.freedesktop.org/airlied/virglrenderer). There we have piglit tests that test ARB_depth_clamp in the virtualized guest and that will rely on the host exposed EXT_depth_clamp. If this is of interest we can also see how to add tests to the OpenGL ES CTS.

dgkoch:
@gerddie I was thinking of the OpenGL ES CTS. 
Unfortunately it doesn't even look like there are any existing GL tests for depth_clamp which could be ported..

PR #242, opened 2019-02-02
Align GLX XML return and param types to upstream specs

nigels-com:
I found some places where the glx.xml seem to be out of sync with the specifications.

The relevant specifications are:

https://www.khronos.org/registry/OpenGL/extensions/SUN/GLX_SUN_get_transparent_index.txt
https://www.khronos.org/registry/OpenGL/extensions/SGIX/GLX_SGIX_pbuffer.txt
https://www.khronos.org/registry/OpenGL/extensions/MESA/GLX_MESA_set_3dfx_mode.txt


oddhack:
Thanks!

PR #243, opened 2019-02-06
Add a private QCOM token

tate-hornbeck:


tate-hornbeck:
We don't plan on publishing this extension, its meant to be exclusively internal/private. Wanted to document the token here so we don't lose track of it.

PR #248, opened 2019-02-20
glx: Add GLX_EXT_context_priority

nwnk:
Copied from the corresponding EGL extension. This does not yet update
the XML for the new enumerants, pending resolution of the issues
mentioned in the extension spec.

### DIFF
--- a/api/GL/glxext.h
+++ b/api/GL/glxext.h
@@ -34,7 +34,7 @@ extern "C" {
 **   https://github.com/KhronosGroup/OpenGL-Registry
 */
 
-#define GLX_GLXEXT_VERSION 20190228
+#define GLX_GLXEXT_VERSION 20190411
 
 /* Generated C header for:
  * API: glx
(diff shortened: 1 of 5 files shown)


pdaniell-nv:
@oddhack do you have any thoughts on the question in issue 1 of this spec?

oddhack:
I think it would be polite to consult with IMG (if you can figure out the right contact), but insofar as the behavior of the GLX EXT enums and the EGL IMG enums is semantically equivalent, it seems safe enough. I don't see why it says "the GLX enumerant registry has assigned that range to IBM" since nothing seems to be currently using the range in glx.xml - am I missing something? If that's a concern Ian Romanick might have some recollection of that from when he worked at IBM (at least, I think he used to).

nwnk:
> I don't see why it says "the GLX enumerant registry has assigned that range to IBM" since nothing seems to be currently using the range in glx.xml - am I missing something?

My apologies, I'm not sure how I got that confused. I think I was looking at the 3000-3999 range for vendor-private opcodes, but that's obviously not the same thing.

I'll ask around for the appropriate contact at IMG.

pdaniell-nv:
@nwnk Any progress on asking around at IMG?

putsman:
Given the precedent of sharing enums between EGL and GLX, Imagination has no problem in reusing the enum values from EGL_IMG_context_priority for a functionally-equivalent GLX extension.

oddhack:
@nwnk is this ready to be merged now? I think all the issues have been addressed.

PR #251, opened 2019-02-25
Adds ES support to NV_scissor_exclusive

pknowlesnv:
Adds ES extension number.

Adds NV_viewport_array and OES_viewport_array interactions for MAX_VIEWPORTS and GetFloati_v.

PR #246, opened 2019-02-19
Add extension EXT_texture_query_lod for GLES

gerddie:
This extension is a rewrite of  ARB_texture_query_lod  to provide this functionality in OpenGL ES.

PR #254, opened 2019-02-28
Update registry and headers

dgkoch:
Also add XML entry for EXT_texture_query_lod
and mark EXT_texture_query_lod and EXT_depth_clamp as complete.

cc @gerddie


dgkoch:
Header and registry updates needed after #246 and #251. Also noted the LOD extension was missing from the XML (and subsequently the header) so I fixed that too.

PR #255, opened 2019-03-14
Add a "Specification Update Policy" to Khronos-approved extensions

oddhack:
Per internal gitlab opengl/API issue 103.


dgkoch:
Looks reasonable to me. My only suggestion might be to say "OpenGL or OpenGL ES Working Group", but I'm not sure it's worth touching all the files again to do that..

pdaniell-nv:
We discussed this PR in the OpenGL/ES meeting today and are happy with the changes.

One non-urgent tweak we would like to see is for the "OES" extensions to refer to the "OpenGL ES working group" rather than the "OpenGL working group", and for the "KHR" extensions to refer to both the "OpenGL and OpenGL ES working groups". Otherwise everything looks good. Thanks.

PR #258, opened 2019-03-26
Remove aliasing of glRenderbufferStorageMultisampleEXT with glRenderb…

JoergArm:
…ufferStorageMultisample

glRenderbufferStorageMultisampleEXT is less restrictive
on multi-sample preservation and may resolve in comparison
to glRenderbufferStorageMultisample.

Hence the functionality of the latter can't be expressed
through the first.

Remove the aliasing from the gl.xml, and explicitly point
this out in the extension spec.

### DIFF
--- a/extensions/EXT/EXT_multisampled_render_to_texture.txt
+++ b/extensions/EXT/EXT_multisampled_render_to_texture.txt
@@ -21,8 +21,8 @@ Status
 
 Version
 
-    Last Modified Date: June 28, 2016
-    Revision: 7
+    Last Modified Date: March 15, 2019
+    Revision: 8
 
 Number
 
@@ -280,6 +280,11 @@ Dependencies on GL and ES profiles, versions, and other extensions
            "The maximum number of samples supported can be determined by calling
            GetInternalformativ with a pname of SAMPLES."
 
+       RenderbufferStorageMultisampleEXT is not an alias of the OpenGL ES 3.0
+       core function RenderbufferStorageMultisample as only the former (described
+       by this extension) allows resolving multisample information in an implicit
+       way.
+
 Errors
 
     The error OUT_OF_MEMORY is generated when 
@@ -529,6 +534,9 @@ Issues
 
 
 Revision History
+
+    Revision 9, 2019/03/15
+     - Clarified relation to OpenGL ES 3.0 RenderbufferStorageMultisample.
 
     Revision 8, 2018/04/11
      - Clarified wording around implicit resolves, and added Issue 7.
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -25485,7 +25485,6 @@ typedef unsigned int GLhandleARB;
             <param group="InternalFormat"><ptype>GLenum</ptype> <name>internalformat</name></param>
             <param><ptype>GLsizei</ptype> <name>width</name></param>
             <param><ptype>GLsizei</ptype> <name>height</name></param>
-            <glx type="render" opcode="4331"/>
         </command>
         <command>
             <proto>void <name>glRenderbufferStorageMultisampleANGLE</name></proto>
@@ -25528,7 +25527,6 @@ typedef unsigned int GLhandleARB;
             <param group="InternalFormat"><ptype>GLenum</ptype> <name>internalformat</name></param>
             <param><ptype>GLsizei</ptype> <name>width</name></param>
             <param><ptype>GLsizei</ptype> <name>height</name></param>
-            <alias name="glRenderbufferStorageMultisample"/>
             <glx type="render" opcode="4331"/>
         </command>
         <command>


pdaniell-nv:
The PR comes from internal issue https://gitlab.khronos.org/opengl/API/issues/110.

oddhack:
LGTM.

ianromanick:
Is this an actual problem?  We've been shipping these as aliased for years without any known issues.  Un-aliasing these will cause a serious ABI break for us, and I'd rather not do that unless there is a very, very compelling reason to do so.

ianromanick:
This is made even more annoying by the existence of `glRenderbufferStorageMultisampleEXT` in OpenGL ES 2.0 via GL_EXT_multisampled_render_to_texture.  If an OpenGL ES 3.0 implementation supports this extension, it seems that `glRenderbufferStorageMultisampleEXT` and `glRenderbufferStorageMultisample` _should_ alias.  Having functions with the same pair of names that alias or don't alias based on the API is a bit of a mess.

JoergArm:
One can use glRenderbufferStorageMultisample instead of glRenderbufferStorageMultisampleEXT, but not the other way around, so those two aren't exact aliases. This has surfaced in a proposed update to the Conformance tests which started to route calls from glRenderbufferStorageMultisample into glRenderbufferStorageMultisampleEXT due to autogenerating code based on the XML.

ianromanick:
Looking at the gitlab issue, it seems that the problem is the ES version of `glRenderbufferStorageMultisampleEXT` from GL_EXT_multisampled_render_to_texture.  This is a mess.  As far as I can tell, in desktop OpenGL, `glRenderbufferStorageMultisampleEXT` and `glRenderbufferStorageMultisample` _should_ alias.

In an ideal world, GL_EXT_multisampled_render_to_texture would have picked a different name for a function that means something different than existing `glRenderbufferStorageMultisample{EXT,}`.

evelikov:
@ianromanick just had a long close look at the existing code - this change should not be an issue on our end ;-)

ianromanick:
This particular patch won't break anything because we are not using the Khronos XML to generate anything for our driver.  However, that is a goal.  The concept of unaliasing these functions _is_ a problem for desktop OpenGL.  The function in question exists on desktop OpenGL via [GL_EXT_framebuffer_multisample](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_framebuffer_multisample.txt).  That extension was published six years before [GL_EXT_multisampled_render_to_texture](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_multisampled_render_to_texture.txt).

The GLX protocol opcode for `glRenderbufferStorageMultisampleEXT` and `glRenderbufferStorageMultisample` is the same.  On implementations of GLX, it is impossible to unalias them for indirect rendering because the server only sees the opcode.  This is clearly defined in [GL_EXT_framebuffer_multisample](https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_framebuffer_multisample.txt).  Removing the GLX protocol opcode from `glRenderbufferStorageMultisampleEXT` in this MR is unarguably incorrect.

Unaliasing functions does break the interface between libGL and the DRI driver.  We had this problem before when we had to unalias some GL_EXT_framebuffer_object and GL_ARB_framebuffer_object functions in Mesa commit 9f07ca11c17 ("mesa: Dispatch ARB_framebuffer_object and EXT_framebuffer_object differently").  From that commit message:

> HOWEVER, this does represent a compatibility break between the loader (libGL or the Xserver GLX module) and the driver.  Mesa drivers compiled without this change will request a single dispatch table entry for glBindFramebuffer and glBindFramebufferEXT.  Since the updated loader has different entries for each, the request will fail, and the driver will die in a fire.

In that case, it was definitely our bug.  We were (correctly) failing conformance, so we had no other choice than to take our lumps.  Hopefully you can understand my reluctance to take someone else's lumps.

I'm just going to say... this is why MRs need more than 2 hours before they get merged.  In other projects that I have experience with that use this sort of workflow, 24 hours is the _minimum_.

JoergArm:
Aight - I was unaware that there's an "equivalent" extension function defined in the GL (non-ES) land.
Trading one issue for another can't the intention either, so a revert seems the best choice before trying to tackle this again.

nwnk:
> On implementations of GLX, it is impossible to unalias them for indirect rendering because the server only sees the opcode.

We only need to unalias them for indirect GLES contexts though, right? Make a new `GLX_EXT_multisampled_render_to_texture` that defines a new opcode for the EXT name in GLES. If the server supports it then you can use that opcode, otherwise you resort to the old encoding and accept the failure mode.

I admit this doesn't address the ABI between libGL and the driver, though an analogous solution there seems straightforward. I'm personally okay with Mesa ceasing to treat that interface as an external ABI promise, but that's not really a discussion for this bug.

ianromanick:
@nwnk The problem this creates for Mesa is in desktop OpenGL which has functions with the same names that must alias.  In both the Khronos gl.xml and our XML, all the functions from both APIs live in a single namespace.  There isn't a way that I know of to have one set of (`glRenderbufferStorageMultisample`, `glRenderbufferStorageMultisampleEXT`) that alias each other and another set of (`glRenderbufferStorageMultisample`, `glRenderbufferStorageMultisampleEXT`) that do not.  Specifically, in Mesa we use the same dispatch table for both APIs.  We have to use the same dispatch table because an application could create both a GL and a GLES context in the same address space.

I'm feeling very painted into a corner here.  :frowning_face:

nwnk:
> The problem this creates for Mesa is in desktop OpenGL which has functions with the same names that must alias.

Can we define "must alias" here? Is the requirement that `GetProcAddress` for these two functions return the same fptr, or that they have the same effect in a GL context? If the latter, I'm not sure why it's an ABI problem for the loader to have two slots for the two entrypoints, presumably dynamically generating the EXT slot. Old drivers might pick up the old static entrypoint for ...EXT and continue to do the wrong thing in indirect GLES contexts, but... okay?

ianromanick:
Presumably the OpenGL ES 2.0 / 3.x CTS isn't ever intentionally calling `glRenderbufferStorageMultisampleEXT`.  As a temporary work-around, could we make the extension handling in the CTS either prefer function names that match exactly or prefer shorter function names?  That should at least restore the old behavior.  That won't solve the problem, but it should allow OpenGL ES 3.x implementations that expose GL_EXT_multisample_render_to_texture to continue passing CTS.

We do still need to find a proper fix.  There's at least one driver in Mesa that exposes this extension, but I don't know what OpenGL ES version it supports.  There has been talk of adding support to a couple other drivers.  At least one of those drivers supports OpenGL ES ≥ 3.0 and OpenGL ≥ 4.2.

oddhack:
Sorry, should have waited on that one. I interpreted the internal issue to be that we should just do it once the PR was ready. I will revert the change if @pdaniell-nv wants.

pdaniell-nv:
@ianromanick do you want to revert? Do you have a proper fix in mind?

ianromanick:
@nwnk "Must" in the sense that if they don't have the same location in the dispatch table, when the X server receives rendering command opcode 4331, it won't know whether to call `glRenderbufferStorageMultisample` or `glRenderbufferStorageMultisampleEXT`.  Since one is from an optional extension and the other is from a fairly high OpenGL version, any particular driver may only support one or the other.

@pdaniell-nv I think we need to revert.  I honestly don't know if a proper, complete fix is possible.

pdaniell-nv:
@oddhack could you revert this merge. Thanks.

oddhack:
See #261 for reversion.

oddhack:
We might want to open a tracking issue on this, even if it will never be closeable, as a way to document the issue (and point to this PR's discussion thread).

PR #260, opened 2019-04-02
Adds ES interactions to NV_shading_rate_image

pknowlesnv:
Allows implementing with ES. If so, removes references to MULTISAMPLE
and double and adds MAX_VIEWPORTS and GetFloati_v alternatives.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -47743,7 +47743,7 @@ typedef unsigned int GLhandleARB;
             </require>
         </extension>
         <extension name="GL_NV_shader_thread_shuffle" supported="gl|glcore"/>
-        <extension name="GL_NV_shading_rate_image" supported="gl|glcore">
+        <extension name="GL_NV_shading_rate_image" supported="gl|glcore|gles2">
             <require>
                 <enum name="GL_SHADING_RATE_IMAGE_NV"/>
                 <enum name="GL_SHADING_RATE_NO_INVOCATIONS_NV"/>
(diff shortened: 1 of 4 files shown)


xml/gl.xml
```diff
@@ -47743,7 +47743,7 @@ typedef unsigned int GLhandleARB;
             </require>
         </extension>
         <extension name="GL_NV_shader_thread_shuffle" supported="gl|glcore"/>
-        <extension name="GL_NV_shading_rate_image" supported="gl|glcore">
+        <extension name="GL_NV_shading_rate_image" supported="gl|glcore|gles2">
```
oddhack: Since there are parts of the extension API that are GL-only, you can mark up the XML to indicate this (and it will result in those APIs only appearing in glext.h, not gl2ext.h). For an example, see the XML for GL_EXT_blend_minmax - the \<require> block is interfaces appearing in all the APIs, while the \<require api="gl"> block is GL-specific interfaces.

pknowlesnv: Thanks for looking at this. I think all functions and enums here can be supported by both GL and ES. Some others such as `glEnable(GL_MULTISAMPLE)`, `glGetDouble` and `glGet*Indexed` aren't in ES but also aren't introduced by the extension.

oddhack:
LGTM. If you'll confirm it's ready to merge and @pdaniell-nv will sign off on it, I'll merge & regenerate headers.

PR #266, opened 2019-04-17
Fix bogus use of enum group "TypeEnum"

oddhack:
I haven't done the archaeology to figure out what happened here. My
guess is that TypeEnum was intended as a catch-all for queriable
enumerant tokens, which are not realistically possible to list, and then
at some point it got used by the BeginConditionalRender commands.

While it isn't a complete fix, I changed the name of this group to
"ConditionalRenderMode" and used it only with those commands, and
removed the tagging of "TypeEnum" for the generic query commands.
Hopefully that won't subvert any outside use of the "TypeEnum" group.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -2636,11 +2636,15 @@ typedef unsigned int GLhandleARB;
             <enum name="GL_TEXTURE31"/>
         </group>
 
-        <group name="TypeEnum">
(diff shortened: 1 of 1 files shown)


joonazan:
Looks good to me. I'm no longer working on the project where I found this, but at least it requires one less workaround now.

PR #269, opened 2019-05-06
Removed duplicate enum values.

jcant0n:
I founded 3 duplicate enum values in groups while I generate a C# bindings.


oddhack:
Thanks, looks good.

PR #271, opened 2019-05-14
Fix noncontiguous extension numbering

oddhack:
At some point the GL non-ARB extension numbering became discontiguous,
probably due to a typo 532 -> 592 in registry.php. This shifts the
extension numbers to remain contiguous. Since the extension numbers
aren't used for anything except imposing an ordering on the list in the
registry index, this should be harmless.


oddhack:
There are also a few cases of trailing whitespace stripped out.

PR #285, opened 2019-06-26
Rename bufSize to count where necessary

pdaniell-nv:
Rename the "bufSize" parameter name to "count" in functions where that parameter refers to an array size rather than a buffer size in bytes. Also rename bufsize to bufSize.

This resolves https://github.com/KhronosGroup/OpenGL-API/issues/52.

Do not merge this PR until the specs have been updated with the new parameter names.


pdaniell-nv:
Discussed this PR in the OpenGL/ES meeting. We also need to match this XML change with API spec changes so the parameter names match. Assigned to @oddhack to make those change. We'll merge this once when API spec side is ready.

oddhack:
@pdaniell-nv the internal spec changes are ready, so it's just a question when we decide to publish an updated API spec, then merge this.

pdaniell-nv:
@oddhack This is approved to merge. Thanks.

PR #287, opened 2019-07-26
Nvx gpu multicast2

rbiermann:
Adding new NVIDIA extensions

GL_NVX_gpu_multicast2
GL_NVX_progress_fence
WGL_NV_multigpu_context

Please review and merge!

Kind regards,
Ralf Biermann (NVIDIA)

PR #288, opened 2019-07-27
Add NVX_gpu_multicast2 (revised)

oddhack:
Replaces #287, since I can't update @rbiermann's branch in-place in their repository. Changes one extension number to avoid collision with recently merged MESA extension. I was unable to clone their branch into this repo and successfully merge it - some sort of commit history incompatibility - so just did a patch against master in this branch, instead. 

@dgkoch @pdaniell-nv I suggest adding the KHR extension to this branch.

Unfortunately, the changes in #287 directly updated the headers rather than adding the new interfaces to gl.xml / wgl.xml. As a result regenerating the headers will not actually include the interfaces for the new extensions. If you can attempt to add the interfaces for the new extensions introduced in #287 that would be great. If not I'll attempt to do it, but I can't guarantee I'll have time by Sunday evening, with the Vulkan update taking up my time. In terms of just publishing the specs, the branch should be fine as is.

### DIFF
--- a/api/GL/glcorearb.h
+++ b/api/GL/glcorearb.h
@@ -3900,6 +3900,22 @@ GLAPI void APIENTRY glMaxShaderCompilerThreadsKHR (GLuint count);
 #define GL_CONTEXT_ROBUST_ACCESS          0x90F3
 #endif /* GL_KHR_robustness */
 
+#ifndef GL_KHR_shader_subgroup
+#define GL_KHR_shader_subgroup 1
+#define GL_SUBGROUP_SIZE_KHR              0x9532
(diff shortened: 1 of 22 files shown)


rbiermann:
I'm fine with #287 , Is something else needed from me?

Thanks.

dgkoch:
> OK. The .(w)gl.xml files contain the ext definitions, is there something else needed?

They add new sections for the extensions with the tokens they use, but the tokens aren't defined.

I'll fix it.

dgkoch:
@oddhack I'd added XML support for the new multicast extensions and added the new KHR and corresponding NV extension to this branch. 

I think this is ready to be merged, if you are happy with it.

rbiermann:
Thanks a lot for the help!

PR #289, opened 2019-08-02
Allow GL_MESA_framebuffer_flip_y as a GLES 3.0 extension

fritzk:
Allow GL_MESA_framebuffer_flip_y to implement glFramebufferParameteri
and glGetFramebufferParameteriv so that this extension can
be used by GLES 3.0.


fritzk:
@chadversary please take a look

ianromanick:
I'm going to double check with the ES working group that we can't just add `glFramebufferParameteri` and `glGetFramebufferParameteriv` to OpenGL ES 3.0 contexts.  I'm pretty sure I know the answer, but it won't hurt to ask.

ianromanick:
So, we're going to have to add the `MESA` decorated functions.  Since this has already shipped in a version of Mesa without these functions, we should add an issue explaining some 19.x versions of Mesa have this extension but lack the `MESA` functions.  I think the window of exposure and the potential set of application using this extension is small enough that this won't be a significant problem.

With that addition, I think this PR should be fine to approve.

fritzk:
Thanks for looking into it.  I was too was hoping to avoid the decorated functions.

extensions/MESA/MESA_framebuffer_flip_y.txt
```diff
@@ -32,7 +32,7 @@ Number
 
 Dependencies
 
-    OpenGL ES 3.1 or OpenGL 4.3 is required, for FramebufferParameteri.
+    OpenGL ES 3.0 or OpenGL 4.3 is required
```
chadversary: Minor nit. Missing a period.

extensions/MESA/MESA_framebuffer_flip_y.txt
```diff
@@ -59,7 +59,10 @@ Issues
 
 New Procedures and Functions
 
-    None
+    If ARB_framebuffer_no_attachments is not implemented then implement
```
chadversary: For ease of use, this extension should implement the below functions independently of the availability of `GL_ARB_framebuffer_no_attachments`. As written now, if a driver currently does not have `GL_ARB_framebuffer_no_attachments`, then the driver must provide the MESA functions; later, if the driver begins roviding `GL_ARB_framebuffer_no_attachments`, then the driver must *stop* providing the MESA functions, potentially breaking apps that use the MESA functions.

extensions/MESA/MESA_framebuffer_flip_y.txt
```diff
@@ -59,7 +59,10 @@ Issues
...
+
+    void FramebufferParameteriMESA(enum target, enum pname, int param);
+    void GetFramebufferParameterivMESA(enum target, enum pname, int *params);
```
chadversary: About the 'New Tokens' section below...

I believe it should read as follows, plus an "dependency interaction" section is needed.


```
New Tokens

    Accepted by the <pname> argument of FramebufferParameteriMESA and
    GetFramebufferParameterivMESA:

        FRAMEBUFFER_FLIP_Y_MESA                      0x8BBB

Dependencies on OpenGL 4.3, OpenGL ES 3.1, ARB_framebuffer_no_attachments, and
any other versions and extensions that provide the entry points FramebufferParameteri
and FramebufferParameteriv

        Token FRAMEBUFFER_FLIP_Y_MESA is accepted as the <pname> argument of
        FramebufferParameteri and FramebufferParameteriv.
```

extensions/MESA/MESA_framebuffer_flip_y.txt
chadversary: Hmm... I have more opinions now :-)

The extension requires GL 4.3 or GLES 3.0. GL 4.3 already provides glFramebufferParameteri. So, you could reasonably change the symbol visibility aspect of the extension to:

  * In GL >= 4.3, no MESA functions. Only core functions.
  * In GL < 4.3, don't care, not supported.
  * In GLES >= 3.0. Always provide MESA functions.
  * In GLES >= 3.1. Support the core functions too.

Whether you choose to follow my first comment (provide MESA functions everywhere) or this comment (provide MESA functions in GLES only), I don't have a strong opinion.

extensions/MESA/MESA_framebuffer_flip_y.txt
chadversary: The *Dependencies* are correct as-is. But you could soften them a bit as:

```
Requires OpenGL ES 3.0, OpenGL 4.3, or ARB_framebuffer_no_attachments.
```

extensions/MESA/MESA_framebuffer_flip_y.txt
fritzk: I think the second option makes sense.

ianromanick:
Looks good to me.

pdaniell-nv:
@oddhack, this one is approved to merge.

PR #291, opened 2019-08-09
Add a warning comment to the <group> tags explaining that they may be out of date

### LINKED ISSUE #280: State Khronos' official stance on enum groupings
Enum groupings are great for downstream consumers of the Khronos specification (i.e. bindings generators for other languages such as C#) and are depended on by projects such as OpenTK. However, Khronos seems to be neglecting these groupings, breaking code generators that depend on them. I feel that Khronos either needs to stop neglecting these groupings or officially state that enum groups are deprecated so that library developers can start looking for alternatives.

Regards,

Dylan.

### LINKED ISSUE #286: Link returns 404 on the "Tutorials and How To Guides" section.
Link to the section:

https://www.khronos.org/opengl/wiki/Getting_Started#Tutorials_and_How_To_Guides

The first link, named, "Develop with OpenGL 3.1 from scratch (Français)" now returns 404 when opened. This content no longer exists.

http://www.siteduzero.com/tutoriel-3-421558-developpez-vos-jeux-video-3d-avec-opengl-3-1.html

oddhack:
Fixes #280
Not in the sense of "making the group tags complete", but in the sense of "explaining that they aren't complete and we would need help in changing that".


xml/gl.xml
```diff
@@ -79,7 +79,14 @@ typedef unsigned int GLhandleARB;
         <type>typedef void (<apientry/> *<name>GLVULKANPROCNV</name>)(void);</type>
     </types>
 
-    <!-- SECTION: GL parameter class type definitions. -->
+    <!-- SECTION: GL parameter class type definitions.
+         The groups are intended to contain all the possible legal values
+         for corresponding function parameters, but it is likely that many
+         of the groups are out of date relative to current OpenGL and OpenGL
+         ES specifications, and the many extensions to those specifications.
```
Perksey: We should make the reliability (or lack thereof) of the groupings more explicit.
```suggestion
         ES specifications, and the many extensions to those specifications.
         As such, they may not be a reliable source for enumeration info.
```
or
```suggestion
         ES specifications, and the many extensions to those specifications.
         You should prefer the GL enumerant definitions where possible.
```

xml/gl.xml
```diff
@@ -79,7 +79,14 @@ typedef unsigned int GLhandleARB;
...
+         We welcome assistance from the community in making the enum groups
+         more complete.
```
Perksey: I think there should be a message explaining that the groupings aren't used by Khronos, and it is unlikely that the groupings will be updated by Khronos themselves.

frederikja163:
You might wanna take a look at #292 before you merge this PR

Perksey:
#292 is a great maintenance PR, but I still think we should keep this warning in the spec as without Khronos' backing of the groupings, it won't be hard for PRs to slip through which degrade the validity of the groupings.

xml/gl.xml
```diff
@@ -79,7 +79,18 @@ typedef unsigned int GLhandleARB;
...
 
-    <!-- SECTION: GL parameter class type definitions. -->
+    <!-- SECTION: GL parameter class type definitions.
+         The groups are intended to contain all the possible legal values
+         for corresponding function parameters, but it is likely that many
+         of the groups are out of date relative to current OpenGL and OpenGL
+         ES specifications, and the many extensions to those specifications.
+         As such, they may not be a reliable source for enumeration info.
+
+         We welcome assistance from the community in making the enum groups
```
fluppeteer: Nit: If #292 currently (maybe) fixes everything, it might be better to say "We welcome assistance from the community in ensuring that the enum groups are complete." That is - we may not be sure that there's anything wrong, so although we should warn that they might not be, and community assistance in checking either way would be useful, we might not want to claim there *is*. It might be embarrassing for a student to make a term project in open source software out of completing these enums and then find that there was nothing to do, for example.

Perksey: There's no guarantee that any of the groupings are complete. Fred's PR only *adds* missing groups, it doesn't at all look at any of the groupings that were already in the registry (which might not contain all possible values).

I don't think the groupings should ever be implied to be complete without Khronos managing them themselves.

Perksey:
Side note: Has the Khronos group investigated moving to the Vulkan specification format? (i.e. merging all enum groupings into `enums` with `namespace`s)

oddhack:
@Perksey the Vulkan/OpenXR schema has some significant differences although it started off with the GL stuff, so it would be a lot of work to switch over. Not currently planned.

Perksey:
Do you have XSD schemas for both? I can try and make an auto migrate
utility to take to the board’s next GL meeting. I’ve already converted the
GL spec into 3 different formats haha

oddhack:
There are RNC schemas for both. Converting the XML file would be a small part of the problem, though.

xml/gl.xml
Perksey: Jon applied the first suggestion without marking the review as resolved ;) As for suggestion 2, while moot, I was referring to the `<enums namespace=` portion of the spec.

pdaniell-nv:
Thanks for the input on this PR so far. Any other comments before we accept this change? Thanks.

xml/gl.xml
```diff
@@ -79,7 +79,18 @@ typedef unsigned int GLhandleARB;
...
+         more complete. Khronos does not use the enum group information
+         ourselves, and the OpenGL Working Group does not have internal
```
Perksey: Grammatical nitpick :|
```suggestion
         themselves, and the OpenGL Working Group does not have internal
```

Perksey:
After that change has been applied, this PR has my green light.

xml/gl.xml
oddhack: I just dropped the word, rather than open up to more potential grammar issues.

xml/gl.xml
fluppeteer: Does my proposal:

> We welcome assistance from the community in ensuring that the enum groups are complete.

...imply that we think they *are*? I just didn't like us saying that they're definitely *incomplete* (which is how I read "making the enum groups more complete"), unless we want to validate this statement alongside the current state of completeness.

Is this better?

> We welcome assistance from the community in achieving and maintaining the completeness of the enum groups.

pdaniell-nv: I like this better.

Perksey: Yeah go with it.

oddhack: Done (hopefully - thread has gotten pretty dense to follow all the suggestions).

xml/gl.xml
oddhack: Done previously.

oddhack:
@pdaniell-nv I think this should be OK now? AFAICT everyone's comments / requests have been incorporated.

pdaniell-nv:
@oddhack yep looks good. Approved to merge.

PR #298, opened 2019-09-09
Add spec language for compute derivatives with variable group sizes.

nvpbrown:
In NV_compute_shader_derivatives, we disallow the "quads" derivative
mode if the local workgroup size has an odd width or height and
disallow the "linear" derivative mode if the number of threads in the
local workgroup is not a multiple of four.  However, the spec didn't
consider interactions with ARB_compute_variable_group_size, which
allows applications to compile shaders without a fixed group size.  We
can't throw errors for bad group sizes in this case.

In discussion on a Khronos GitHub issue, we decided to instead
implement this check at run-time and throw INVALID_VALUE if a bad
group size is passed to a dispatch call.

This change updates the NV_compute_shader_derivatives spec to specify
a run-time check with no compile-time check in this case.


pdaniell-nv:
Vendor extension bug fix approved to merge. @oddhack

PR #299, opened 2019-09-11
Update copyright headers

Perksey:
Just updates the year in the XML headers because my OCD eye noticed it.


Perksey:
Should be go for merge with zero damage.

oddhack:
Looks fine, thanks for catching that. @pdaniell-nv please mark with Milestone Approved when you're OK with it.

PR #300, opened 2019-09-18
Stop treating 'effective' values for DSA functions as passed arguments

oddhack:
for the purposes of error code generation (matching earlier changes to the GL 4.6 spec). Fixes internal API issue 22.


pdaniell-nv:
Appears to be a faithful reflection if the spec.

PR #301, opened 2019-09-18
Prevent user interface variables from containing a value of type OpTypeBool

oddhack:
Closes internal API issue 111.


pdaniell-nv:
Looks good, thanks.

PR #307, opened 2019-09-30
Add OES enums to groups

tsuoranta:
Some OES enums are missing groups. Fix for #308 
Also fixes duplicate `GL_FRAMEBUFFER_ATTACHMENT_OBJECT_NAME` enum in `FramebufferAttachmentParameterName` group and replaces the second instance with `GL_FRAMEBUFFER_ATTACHMENT_OBJECT_TYPE` enum which it probably should be.


pdaniell-nv:
@oddhack if you are happy with this its approved to merge.

PR #316, opened 2019-10-12
[gl.xml] Add InvalidateFramebufferAttachment enum group

tsuoranta:
Add a variation of `FramebufferAttachment` enum group for `glDiscardFramebuffer()` and `glInvalidate*Framebuffer()` attachments which also accepts `GL_COLOR`, `GL_DEPTH` and `GL_STENCIL` attachments.

Fix proposal for #317


oddhack:
LGTM. @pdaniell-nv please marked Approved when you're OK with it.

This is where the 'group' concept begins to show real weakness - there are a lot of GL calls that accept *almost* the same set of valid enum values.

PR #319, opened 2019-10-13
regenerate GLES3 & GLSC2 headers

1ace:
These were forgotten in dd9888c1ad20a1e8c502 ("Regenerated headers with
new format header guards that don't prevent").

### DIFF
--- a/api/GLES3/gl31.h
+++ b/api/GLES3/gl31.h
@@ -1,12 +1,12 @@
-#ifndef __gl31_h_
-#define __gl31_h_ 1
+#ifndef __gles2_gl31_h_
+#define __gles2_gl31_h_ 1
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 /*
-** Copyright (c) 2013-2016 The Khronos Group Inc.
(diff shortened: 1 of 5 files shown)


api/GLES3/gl31.h
```diff
@@ -1,12 +1,12 @@
-#ifndef __gl31_h_
-#define __gl31_h_ 1
+#ifndef __gles2_gl31_h_
```
oddhack: Did you re-run genheaders.py against these targets specifically, or was this a manual edit? They were taken out from the Makefile GLHEADERS target at some point, probably because they are out of date versions of the API, but no reason not to regenerate them. The edits seem fine to me but I'd like to get signoff from @pdaniell-nv before merging.

pdaniell-nv:
OpenGL/ES working groups approve this PR.

api/GLES3/gl31.h
oddhack: Huh, bizarre. gl31.h isn't one of the targets, AFAICT. But harmless.

PR #322, opened 2019-10-30
Fix "shared p-bits" specification in ARB_texture_compression_bptc

### LINKED ISSUE #310: Possible error in ARB_texture_compression_bptc document
According to the ARB_texture_compression_bptc extension document, in modes with shared P-bits, there is one P-bit per endpoint (https://github.com/KhronosGroup/OpenGL-Registry/blob/master/extensions/ARB/ARB_texture_compression_bptc.txt#L372-L374):
```
[...] If the format has a
shared P-bit, there are two bits for endpoints 0 and 1 from low to
high. [...]
```
However, in Microsoft's documentation of the DirectX BC7 compression method, which I understand to be the same as BPTC (please correct me if not), there is instead one P-bit per subset (https://docs.microsoft.com/en-us/windows/win32/direct3d11/bc7-format-mode-reference):
```
RGBP 6.6.6.1 endpoints with a shared P-bit per subset)
```
Microsoft's reference BC7 decoder also treats shared P-bits as per subset, not per endpoint (https://github.com/microsoft/DirectXTex/blob/master/DirectXTex/BC6HBC7.cpp#L3293-L3301):
```
// shared pbits
for (uint8_t ch = 0; ch < BC7_NUM_CHANNELS; ch++)
{
    for (size_t i = 0; i <= uPartitions; i++)
    {
        pFixedEndpoints[i].A[ch] = static_cast<uint8_t>((pFixedEndpoints[i].A[ch] << 1) | pbits[i]);
        pFixedEndpoints[i].B[ch] = static_cast<uint8_t>((pFixedEndpoints[i].B[ch] << 1) | pbits[i]);
    }
}
```
Coincidentally, the only mode which uses shared P-bits (mode 1) has 2 subsets, so this difference in interpretations does not change the number of P-bits and thus does not affect the decoding of the rest of the block, though it does result in differing low-order bits of the resulting color values.

Is this an intentional difference between OpenGL BPTC and DirectX BC7, or is the BPTC extension document incorrectly worded?

pdaniell-nv:
As discussed in issue #310, the "shared p-bits" specification in the ARB_texture_compression_bptc extension did not match what's in the Khronos Data Format spec (https://www.khronos.org/registry/DataFormat/specs/1.3/dataformat.1.3.html) or the DX spec.

This PR fixes that and closes #310.

### DIFF
--- a/extensions/ARB/ARB_texture_compression_bptc.txt
+++ b/extensions/ARB/ARB_texture_compression_bptc.txt
@@ -46,8 +46,8 @@ Status
 
 Version
 
-    Last Modified Date:         June 6, 2016
-    Revision:                   8
+    Last Modified Date:         October 30, 2019
+    Revision:                   9
 
 Number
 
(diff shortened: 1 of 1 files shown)


pdaniell-nv:
Approved to merge.

lexaknyazev:
Please apply this update to the `EXT_texture_compression_bptc` (OpenGL ES version of the same extension).

pdaniell-nv:
Fox the EXT_texture_compression_bptc is here: https://github.com/KhronosGroup/OpenGL-Registry/pull/337

PR #324, opened 2019-11-05
[gl.xml] Fix glMultiDrawArrays() param len expression

tsuoranta:
`glMultiDrawArrays` had incorrect param len for `first` parameter.

Fix for #325


pdaniell-nv:
@oddhack Approved.

Issue #326, opened 2019-11-12
EXT_disjoint_timer_query: gl.xml missing glGetInteger64vEXT


ascent12:
https://www.khronos.org/registry/OpenGL/extensions/EXT/EXT_disjoint_timer_query.txt mentions in the "Dependencies" section that

>Interaction: This extension adds GetInteger64vEXT if OpenGL ES 3.0 is not supported

However, there is no specification of this function anywhere in gl.xml or in the final gl2ext.h header.
Without this, an application targeting GLES2 that wants to use this extension cannot query GL_TIMESTAMP_EXT to its full precision if GL_QUERY_COUNTER_BITS_EXT is greater than 32, e.g. Mesa on Linux.

oddhack:
It would not be hard to inject this into gl.xml, but it's worth asking first if this is a real-world use case -  *is* there a GLES2 implementation supporting this extension that isn't also a GLES3 implementation at the same time? If so you're welcome to construct a PR against gl.xml to alias the core function and add a gles2-specific requirement for it in the extension interface block.

tpalli:
I've created a PR here:
https://github.com/KhronosGroup/OpenGL-Registry/pull/329

tpalli:
@oddhack what comes to the GLES2 driver question, I know that we have drivers in Mesa that support only GLES2 (+ possibly subset of GLES3.x features), this could be useful for such a driver.

oddhack:
I overlooked that the extension specification itself will also need some text added to define the GetInteger64vEXT command, since that's present nowhere else in the API. That could probably most easily be cloned from the GL_APPLE_sync spec since they did a similar thing, although I haven't verified that's identical to the language for GetInteger64v in the ES 3.2 spec.

oddhack:
Added #338 with the spec language for GetInteger64vEXT.

PR #329, opened 2019-11-19
[gl.xml] add glGetInteger64vEXT to EXT_disjoint_timer_query

tpalli:
From EXT_disjoint_timer_query spec:
   "Interaction: This extension adds GetInteger64vEXT if
    OpenGL ES 3.0 is not supported"

Signed-off-by: Tapani Pälli <tapani.palli@intel.com>

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -18216,6 +18216,12 @@ typedef unsigned int GLhandleARB;
             <alias name="glGetInteger64v"/>
         </command>
         <command>
+            <proto>void <name>glGetInteger64vEXT</name></proto>
+            <param group="GetPName"><ptype>GLenum</ptype> <name>pname</name></param>
+            <param len="COMPSIZE(pname)"><ptype>GLint64</ptype> *<name>data</name></param>
+            <alias name="glGetInteger64v"/>
+        </command>
+        <command>
             <proto>void <name>glGetIntegerIndexedvEXT</name></proto>
             <param><ptype>GLenum</ptype> <name>target</name></param>
             <param><ptype>GLuint</ptype> <name>index</name></param>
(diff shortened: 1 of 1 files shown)


xml/gl.xml
```diff
@@ -45003,6 +45009,7 @@ typedef unsigned int GLhandleARB;
                 <command name="glGetQueryObjectuivEXT"/>
                 <command name="glGetQueryObjecti64vEXT"/>
                 <command name="glGetQueryObjectui64vEXT"/>
+		<command name="glGetInteger64vEXT"/>
```
dgkoch: hard tabs?

tpalli: sorry, fixed that

oddhack:
@tpalli also see my comment https://github.com/KhronosGroup/OpenGL-Registry/issues/326#issuecomment-555544267 .

@dgkoch @pdaniell-nv I'm OK with this in principle as it seems to just do what the spec already required, but the extension was incompletely specified so it's worth asking in the WG if anyone has concerned about this. The net effect will be to add a new GetInteger64vEXT entry point in gl2ext.h, which should be implemented by anyone shipping this extension.

jackohound:
Qualcomm has some interest in this fix.  Beyond adding GetInteger64vEXT to the gl2ext.h, the extension text should should be updated to document the GetInteger64vEXT entrypoint.

Also, we'd like to add an Issue to the extension, documenting that earlier versions of this extension did not formally add GetInteger64vEXT.  Because of that, ES3.0+ driver implementations may exist that expose GetInteger64v but not GetInteger64vEXT.  This extension currently states that drivers should expose both.

We can draft these changes as a separate PR.

pdaniell-nv:
Since the spec fix will be in another PR, this PR for the header fix is approved to merge.

PR #337, opened 2019-12-10
Fix "shared p-bits" specification in EXT_texture_compression_bptc

pdaniell-nv:
This is a follow-on to https://github.com/KhronosGroup/OpenGL-Registry/pull/322, which fixed the "shared p-bits" for ARB_texture_compression_bptc. This makes the same fix to the OpenGL ES EXT_texture_compression_bptc.


pdaniell-nv:
Approved to merge.

PR #338, opened 2019-12-11
Add specification language for GetInteger64v to EXT_disjoint_timer_query

### LINKED ISSUE #326: EXT_disjoint_timer_query: gl.xml missing glGetInteger64vEXT

oddhack:
Also add an Interactions section specifying that it is only required when only OpenGL ES 2.0 is
supported.

Closes #326

### DIFF
--- a/extensions/EXT/EXT_disjoint_timer_query.txt
+++ b/extensions/EXT/EXT_disjoint_timer_query.txt
@@ -250,10 +250,10 @@ Additions to Chapter 5 of the OpenGL ES 2.0 Specification (Special Functions)
 Additions to Chapter 6 of the OpenGL ES 2.0 Specification (State and State
 Requests)
 
-    Add GetInteger64v to section 6.1.1 "Simple Queries", following
+    Add GetInteger64vEXT to section 6.1.1 "Simple Queries", following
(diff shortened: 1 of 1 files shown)


oddhack:
@pdaniell-nv this pulls in the spec language for the GetInteger64v query added to the XML in #329. Should be a no-brainer although considering the number of typos I made not including EXT after 64v, some more review would be welcome :-)

pdaniell-nv:
@jackohound Can you review this?

jackohound:
LGTM.  Thanks, @oddhack for making these changes.

pdaniell-nv:
Thanks @jackohound. Approved to merge.

Issue #335, opened 2019-12-06
XML Format Consultation: Merging <groups> and <enums>


Perksey:
# Introduction
One thing that's annoyed me about the current XML schema is that the strongly-typed enum groupings are separate to the rest of the specification. Granted, they aren't that useful in C but downstream consumers (such as my [Silk.NET](https://github.com/Ultz/Silk.NET) library for C#) who are using the XML specification to bind OpenGL to languages that do better support strongly typed enums.

# Today's problem
The problem with the group tags being separate is that there's a lesser emphasis on the correctness of the groupings. And, to be fair to the Khronos Group, I see why they wouldn't need to worry about the groupings as they don't use them themselves and, as such, won't want to dedicate cycles to ensuring the groupings' correctness.

I want to tackle this problem by merging the `<enum>` blocks and the `<group>` blocks. If these structures are merged, it encourages spec editors to group the enums properly upon creating a new extension, and should hopefully solve the problem of invalid groupings going forward because if creating proper enum groups is a requirement upon vendors and extension writers, we won't have any group problems going forward.

# Solution in action
My proposed solution uses already established schema elements such that it wouldn't be too breaking for the Khronos Group themselves, but may affect downstream consumers. This is why I want to consult with these consumers so that we can establish whether this is a worthy enhancement to the spec.

## Current syntax
```xml
<groups>
    <group name="GroupName">
        <enum name="GL_ENUM_NAME" />
    </group>
    <group name="AnotherGroupAMD">
        <enum name="GL_ANOTHER_ENUM_AMD" />
    </group>
    <group name="AGroupThatReusesTokensEXT">
        <enum name="GL_ENUM_NAME" />
        <enum name="GL_ANOTHER_ENUM_AMD" />
    </group>
</groups>
```

## Proposed Syntax

```xml
<enums namespace="GL" start="0x0001" end="0x0002">
    <enum value="0x0001" name="GL_ENUM_NAME" group="GroupName,AGroupThatReusesTokensEXT" />
    <enum value="0x0002" name="GL_ANOTHER_ENUM_AMD" group="AnotherGroupAMD,AGroupThatReusesTokensEXT" />
</enums>
```

## Pros
- This syntax doesn't require multiple enum definitions for reuse of an enum in multiple groups. (we will use a comma to separate groups in the group element i.e. `group="FirstGroup,SecondGroupINTEL"`)
[...]

PR #343, opened 2020-01-08
Implement phase one of the migration to group attributes

Perksey:
Implements the new schema from #335 and deprecates the old one. That issue should be kept open until the Khronos Group decides to remove the old group blocks permenantly. 

The enumerants have been tagged using an automated process. As such, the groups will have been preserved in this new format exactly as they were in the old format.

The RNC schema has also been updated.


Perksey:
This is ready for review.

xml/registry.rnc
```diff
@@ -62,26 +61,9 @@ Type = element type {
     text
 }
 
-# <groups> defines a group of enum groups
-Groups = element groups {
```
oddhack: I wouldn't think the updated XML would validate against the schema with this stuff removed, while the group tags themselves are marked deprecated but still present?

Perksey: Fixed.

oddhack:
@pdaniell-nv I think this is OK to merge per #335, let me know when WG signs off. Also see #346.

pdaniell-nv:
I'll put this on the agenda for the next OpenGL/ES WG meeting January 22nd. If that is too far out let me know and I'll call an email vote.

Perksey:
January 22nd is good for me, I look forward to getting this signed off.

pdaniell-nv:
Approved to merge. Thanks for your patience.

oddhack:
@Perksey a stray BOM crept into gl.xml in this PR which the vk-gl-cts project just got affected by. Propose to pull it out in #506. I assume this was an accident as it's not mentioned anywhere in the comments above but in any event, the file is defined to be encoded in UTF-8. Probably the Python script should be tweaked to open the registry in UTF-8 mode, although it just contains ASCII.

Issue #341, opened 2020-01-06
EXT_texture_sRGB_decode assumes S3TC sRGB formats introduces in EXT_texture_compression_s3tc


hanfling:
EXT_texture_compression_s3tc only introduces the non sRGB S3TC formats, where as the EXT_texture_sRGB extension actually introduces the COMPRESSED_SRGB_S3TC_DXT1_EXT, COMPRESSED_SRGB_ALPHA_S3TC_DXT1_EXT, COMPRESSED_SRGB_ALPHA_S3TC_DXT3_EXT, and COMPRESSED_SRGB_ALPHA_S3TC_DXT5_EXT formats.

Now inside EXT_texture_sRGB_decode spec:

	Dependencies on EXT_texture_compression_s3tc

		If EXT_texture_compression_s3tc is NOT supported, delete
		COMPRESSED_SRGB_S3TC_DXT1_EXT, COMPRESSED_SRGB_ALPHA_S3TC_DXT1_EXT,
		COMPRESSED_SRGB_ALPHA_S3TC_DXT3_EXT, and
		COMPRESSED_SRGB_ALPHA_S3TC_DXT5_EXT from Section 3.8.15.

If I understand this correctly, this needs to be:

		If EXT_texture_compression_s3tc or EXT_texture_sRGB is NOT supported...

Apple implementation seems to expose EXT_texture_sRGB_decode and EXT_texture_compression_s3tc, but not EXT_texture_sRGB.

pdaniell-nv:
Discussed in the OpenGL/ES joint working group meeting and we agree with your understanding. Would you like to propose a PR to fix the extension?

hanfling:
Assuming PR==pull request. If needed I can do that, but I prefer not to. I'm not too familiar with how to exactly phrase it for the spec/ext.

pdaniell-nv:
> Assuming PR==pull request. If needed I can do that, but I prefer not to. I'm not too familiar with how to exactly phrase it for the spec/ext.

No worries. How does this look: https://github.com/KhronosGroup/OpenGL-Registry/pull/344

hanfling:
The

	Dependencies on EXT_texture_compression_s3tc

but inside it discusses the dependency on both exts, I think it rather needs to be either

	Dependencies on EXT_texture_compression_s3tc and EXT_texture_sRGB

Or have two sections:


	Dependencies on EXT_texture_compression_s3tc

and

	Dependencies on EXT_texture_sRGB

[...]

PR #344, opened 2020-01-08
Fix EXT_texture_compression_s3tc dependency in EXT_texture_sRGB_decode

pdaniell-nv:
This fixes issue #341

### DIFF
--- a/extensions/EXT/EXT_texture_sRGB_decode.txt
+++ b/extensions/EXT/EXT_texture_sRGB_decode.txt
@@ -31,8 +31,8 @@ Status
 
 Version
 
-    Date: November 8, 2017
-    Revision: 0.91
+    Date: January 8, 2020
+    Revision: 0.92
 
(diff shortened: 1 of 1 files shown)


hanfling:
This looks fine to me as far as S3TC is concerned, but I think Section 3.8.15 also misses listing the ETC2/ASTC texture formats, but thats certainly scope of another ticket/pull request.

oddhack:
LGTM. Mark Approved when ready to merge.

pdaniell-nv:
We discussed this PR in the OpenGL/ES meeting today and we're not sure if this change is even needed. I'll follow up in the #341 issue.

pdaniell-nv:
I think this change is not needed as the extension already has a dependency on GL_EXT_texture_srgb, or OpenGL 2.1, where GL_EXT_texture_srgb was incorporated in the core.

PR #348, opened 2020-01-17
Fix typos in S3TC

lexaknyazev:
These parts define _editing_ of compressed textures and refer to offsets, so `*SubImage*` functions should be used.


pdaniell-nv:
Approved to merge. I double-checked this doesn't cause a spec regression by accidentally removing the description for CompressedTexImage2D, which it does not since that's covered further up.

PR #352, opened 2020-01-22
Add missing interactions between ARB_viewport_array and NV_depth_buffer_float

### LINKED ISSUE #350: DepthRangeArraydvNV and DepthRangeIndexeddNV not defined in XML
The interaction with `NV_depth_buffer_float` of `ARB_viewport_array` is not defined in [gl.xml](https://raw.githubusercontent.com/KhronosGroup/OpenGL-Registry/master/xml/gl.xml). The functions added by this interaction are `DepthRangeArraydvNV` and `DepthRangeIndexeddNV`.

The extension [specification](https://www.khronos.org/registry/OpenGL/extensions/ARB/ARB_viewport_array.txt) defines these functions below `Interactions with NV_depth_buffer_float`.

Edit: Issue redirected from glad [here](https://github.com/Dav1dde/glad/issues/242).

Perksey:
# Related Issues
Closes #350 

# Thoughts
- It feels weird putting NVIDIA functions in an ARB extension. Perhaps this isn't the right place for them.
- ARB_viewport_array is supported on both the core and compatibility profiles, whereas NV_depth_buffer_float is only supported on the compatibility profile. I'm not sure whether it's appropriate for compatibility-only functions to be referenced from a core extension

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -48038,7 +48038,7 @@ typedef unsigned int GLhandleARB;
                 <enum name="GL_MAX_DEEP_3D_TEXTURE_DEPTH_NV"/>
             </require>
         </extension>
-        <extension name="GL_NV_depth_buffer_float" supported="gl">
+        <extension name="GL_NV_depth_buffer_float" supported="gl|glcore">
             <require>
                 <enum name="GL_DEPTH_COMPONENT32F_NV"/>
                 <enum name="GL_DEPTH32F_STENCIL8_NV"/>
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -15432,6 +15432,12 @@ typedef unsigned int GLhandleARB;
             <glx type="render" opcode="174"/>
         </command>
         <command>
+            <proto>void <name>glDepthRangeArraydvNV</name></proto>
+            <param><ptype>GLuint</ptype> <name>first</name></param>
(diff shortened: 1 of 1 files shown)


oddhack:
The XML says NV_depth_buffer_float is supported only in the compatibility profile, unless I'm misreading something. Probably it is supported equally well in core and the XML was never updated to express that, since GL 3.2 came out several years after the extension was published. I would be unsurprised to find that's true of other vendor extensions as well.

Perksey:
Yeah compatibility that's what i meant sorry.

Perksey:
What's the proposed course of action? Should the extension be updated to support `gl` and `glcore`, or should we just leave it for now? Also what did you think about my other comment?

oddhack:
I suggest putting the interaction in the NV extension rather than the ARB extension, and perhaps @pdaniell-nv can confirm if NV_depth_buffer_float is supported in core profile.

pdaniell-nv:
I think adding these under the ARB spec is fine, since that's actually where they come from. So the PR looks good to me. And yes, NV_depth_buffer_float is supported equally in core and compatibility, so we should also add glcore to that extension.

Perksey:
Done.

pdaniell-nv:
Thanks. Approved to merge.

PR #366, opened 2020-02-02
Convert registry.py to registy.json

tsuoranta:
It would be nicer to access registry for other uses if it was data driven, for example, formatted as .json.


Perksey:
I'm pretty sure registry.py is used to generate the C headers, so it's unlikely that this will happen. However, a JSON output could be added to registry.py, but this will likely need to be done by a passionate member of the community.

tsuoranta:
> I'm pretty sure registry.py is used to generate the C headers, so it's unlikely that this will happen. However, a JSON output could be added to registry.py, but this will likely need to be done by a passionate member of the community.

Are you sure about that? I think only the xml registry is needed for the header generation. The xml registry references most, but not all extensions, specifically it does not reference some GLSL only extensions.

In any case, if it so happens that `registry.py` is in fact used for this, it is easy to either fix all those to read the JSON instead (ideally), or to add `registry.py` for this kind of purposes, implementation can read registry from the JSON.

Perksey:
Ah yeah my bad, registry.py is used to feed genheaders.py information. I think the only data-driven source of information will be the XML registry, I doubt Khronos will find any other solution desirable.

oddhack:
This isn't going to happen. The community is encouraged to write whatever tools it wants to manipulate the registry, and many have already done so.

Perksey:
+1 agree with oddhack.

PR #381, opened 2020-03-19
Add QCOM_shading_rate

jackohound:
This vendor extension exposes draw-state controlled Variable Rate Shading in OpenGL ES.  The capabilities exposed and guarantees provided are specifically tuned to current generation Adreno GPUs.


xml/gl.xml
```diff
@@ -50972,6 +50987,19 @@ typedef unsigned int GLhandleARB;
         </extension>
         <extension name="GL_QCOM_shader_framebuffer_fetch_rate" supported="gles2">
         </extension>
+        <extension name="GL_QCOM_shading_rate" supported="gles2">
+            <require>
+                <enum name="GL_SHADING_RATE_QCOM"/>
+                <enum name="GL_PRESERVE_SHADING_RATE_ASPECT_RATIO_QCOM"/>
+                <enum name="GL_SHADING_RATE_1X1_PIXELS_QCOM" group="ShadingRateQCOM"/>
```
Perksey: Thanks for striving to add the group attributes, you are a pioneer of the new group system! However, the group attributes are in the wrong place and should instead be declared within the `<enums>` tags, as demonstrated below:
```xml
<enum value="0x96A6" name="GL_SHADING_RATE_1X1_PIXELS_QCOM" group="ShadingRateQCOM" />
```

jackohound: Sure, I will update that.  Some of the other added enums in this extension are intended to be used with existing API functions.  For example, GL_PRESERVE_SHADING_RATE_ASPECT_RATIO_QCOM can be used with glEnable().  Should such enums be tagged with any particular group?

Perksey: Ah yeah glEnable parameters have the group `EnableCap` :)

pdaniell-nv:
@oddhack approved to merge if you're happy with the mechanics, etc.

oddhack:
Merging. @jackohound as a reminder in the future, please don't regenerate the headers when adding a new extension. I will take care of it; otherwise it creates conflicts with any other pending MRs touching the headers.

PR #384, opened 2020-03-26
Fix of typo in EXT_external_objects

hikiko:
Renamed VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT to
VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL which is the valid Vulkan
image layout.


cubanismo:
This is not a typo, it was done on purpose to keep the enum length shorter.  Note all the other _OPTIMAL Vulkan rnums were shortened in GL as well.

PR #393, opened 2020-04-22
ARB_texture_query_lod: update OpenGL version needed

infapi00:
The original requirement of OpenGL version 2.0 and OpenGL Shading
language version 1.30 was inconsistent, as it is not possible a driver
exposing both is not possible. Updating the OpenGL version required.

This fixes the internal API issue 124.


pdaniell-nv:
@oddhack approved to merge.

PR #402, opened 2020-04-29
Strong VertexAttribPointer types

tsuoranta:
glVertexAttribIPointer() to use VertexAttribIType enum group for
type parameter.

glVertexAttribLPointer to use VertexAttribLType enum group for
type parameter.


tsuoranta:
For those who actually use these enum groups, this can potentially cause some minor breakage. IMHO it is plausible and sensible to use the generic vertex attribute type enum for vertex attribute data structures in user code. The strong type enum is only needed for the API calls. Despite this, I do prefer to have the group match the API call, and I am myself willing to pay the cost of having to add static cast to the user code doing the API call.

Perksey:
Due to the groups not being officially supported as is, I believe we can be more relaxed about such changes to the groups and there have been many similar changes in the past.

PR #412, opened 2020-07-14
fix issue https://github.com/KhronosGroup/OpenGL-API/issues/72

devel0:
this patch adds GL_DEPTH_ATTACHMENT to FramebufferAttachment group in order to allow `glFramebufferRenderbuffer` find right this enum.


pdaniell-nv:
@oddhack, if you're happy with this can you merge. Thanks.

PR #414, opened 2020-07-21
Fix parameter ref for glMultiDrawElementsBaseVertexEXT

jvbsl:
Renamed 'primcount' to 'drawcount' matching the naming in glMultiDrawElementsBaseVertex and fixing the referenced parameter in len argument as well, which used 'drawcount' already.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -22597,7 +22597,7 @@ typedef unsigned int GLhandleARB;
             <param len="COMPSIZE(drawcount)">const <ptype>GLsizei</ptype> *<name>count</name></param>
             <param group="DrawElementsType"><ptype>GLenum</ptype> <name>type</name></param>
             <param len="COMPSIZE(drawcount)">const void *const*<name>indices</name></param>
-            <param><ptype>GLsizei</ptype> <name>primcount</name></param>
+            <param><ptype>GLsizei</ptype> <name>drawcount</name></param>
             <param len="COMPSIZE(drawcount)">const <ptype>GLint</ptype> *<name>basevertex</name></param>
             <alias name="glMultiDrawElementsBaseVertex"/>
         </command>


Perksey:
If I remember correctly the rename to `primcount` was intentional when speccing the extension as the original parameter name was `instancecount`, but was changed per `gl4`. If the len attribute referenced the incorrect parameter name then that's probably what should be changed.

frederikja163:
> If I remember correctly the rename to `primcount` was intentional when speccing the extension as the original parameter name was `instancecount`, but was changed per gl4. If the len attribute referenced the incorrect parameter name then that's probably what should be changed.

The methods just above in the xml spec have the same name but their parameters are called drawcount, so i would assume this one is to be called drawcount too

Perksey:
`drawcount` isn't referenced anywhere in the original extension specification. I think, to adhere to it, the parameter name should continue to be `primcount` and the `len` attributes changed to reference the correct parameter. Changes to XML that affect the C headers should be done as sparingly as possible.

Lokathor:
As I understand it though, the name of a function argument doesn't affect the type of the function pointer at all. Regardless of name, the ABI of calling the function will be the same, right?

Perksey:
Likely yeah, however we'd have to change a lot more things in order to change the parameter name (i.e. regenerate headers, update specs, get them approved etc) whereas keeping the `primcount` name is a lot easier and imho less ambiguous as to what constitutes a "draw" to this function. `primcount` was what was specified originally, it's just someone seemingly got all of the `len` attributes incorrect :P

Lokathor:
Sure, I don't think we need to change the name.

I was more concerned because I'm a Rust user who maintains a GL bindings generator and was momentarily confused about the possibility of an ABI issue.

Perksey:
Ah okay, yeah it doesn't affect most ABIs.

pdaniell-nv:

oddhack:
I am a bit confused by the discussion since there seemed to be general agreement not to change the name but the MR still changes it. That said if it's approved that's fine, but the corresponding prototype in the EXT_draw_elements_base_vertex.txt extension spec document should be changed to match - @jvbsl can you add that change to the PR before I merge it?

jvbsl:
I'm not sure what I should do now, because what @Perksey said seemed pretty reasonable to me and would be a less intrusive as well.
So if you could just assure me again I'll get to it in the coming days

oddhack:
One of two things should happen:

- Change the XML and the extension text spec to use the same new parameter name
- Don't change either of them

The headers would change to match after being regenerated, and there aren't any other documents I'm aware of that are relevant. @pdaniell-nv you should decide (or the WG should) what to do. It's an EXT so this isn't a big deal either way IMO.

Perksey:
Oh apparently I was reading the wrong extension specification, this looks good to merge.

jvbsl:
https://github.com/KhronosGroup/OpenGL-Registry/blob/master/extensions/EXT/EXT_draw_elements_base_vertex.txt

That one should be the extension. Where I'm not sure how to read it correctly but it references "primcount", "drawcount" and even still "instancecount" could be that it contains the history as well? I don't know

Perksey:
Hmmm okay apparently I wasn't looking at the wrong one haha, so discard my last comment. Yeah a fix up for the len attributes is best here, and probably better as I think they're incorrect anyway - a COMPSIZE isn't required for a single parameter afaik

pdaniell-nv:
IMO I think we should take this fix as is and also tweak the GL_EXT_draw_elements_base_vertex extension to use "drawcount" instead of "primcount". Funnily enough the extension spec uses "primcount" in the "New Procedures and Functions" section (line 137) and "drawcount" in the body of the spec (line 205).

pdaniell-nv:
I made a new PR with your commit and an additional change for the extension spec. I wanted to update this PR but either my git skills were lacking or I don't have permissions to do that.

The new PR is here: https://github.com/KhronosGroup/OpenGL-Registry/pull/419

jvbsl:
Thank you very much.
Fyi: You should be able to push directly to the branch if you are a maintainer of this repository.(And if the PR requester has "Allow edits by maintainers" activated on the right side of the PR - which I have, afaik it is default). If not there is nothing you can do except perhaps doing a PR on the forked repository, which sometimes is a bad idea if the owner takes too long to merge it^^

PR #420, opened 2020-08-13
Use SPDX License Identifiers instead of including the full Apache / MIT license

oddhack:
text in files. Also implements one recent tweak to the XML into
the generated headers. Otherwise a no-op.

@pdaniell-nv recommend approving ASAP. This has benefits downstream when doing REUSE license checking in repositories which host these files.


pdaniell-nv:
@oddhack, approved to merge.

PR #424, opened 2020-08-23
Fix inconsistent length out and length for NV_command_list

jvbsl:
Length for parameters which use a pointer as an ouptut parameter should have a length of 1 specified. For the fixes here some vendor counterparts do have the length while others don't.

Length for glListDrawCommandsStatesClientNV parameters was not specified despite it being described in the extension spec:
https://www.khronos.org/registry/OpenGL/extensions/NV/NV_command_list.txt

> The <indirects> pointer should point to a list of size <count> of pointers, each of which should point to a command sequence. 

The other parameters can be seen to be used with the same length in description of the DrawCommandsStatesNV function.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -21757,9 +21757,9 @@ typedef unsigned int GLhandleARB;
             <param><ptype>GLuint</ptype> <name>list</name></param>
             <param><ptype>GLuint</ptype> <name>segment</name></param>
             <param len="count">const void **<name>indirects</name></param>
-            <param>const <ptype>GLsizei</ptype> *<name>sizes</name></param>
-            <param>const <ptype>GLuint</ptype> *<name>states</name></param>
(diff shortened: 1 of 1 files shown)


Perksey (approved):
Seems fine. Note that the length attribute is not required (see also the [rnc spec](https://github.com/KhronosGroup/OpenGL-Registry/blob/master/xml/registry.rnc#L151)), and I believe it's not used by Khronos today.

jvbsl:
Yes we know, but it is nice to have it, therefore the PR as it is afaik without any detriments^^

SunSerega (approved):
`len` attribute already allows to remove >400 auto-generated overloads in my bindings without manual review, so thanks from me for supporting it. It really is nice to have.

Perksey:
> Yes we know, but it is nice to have it, therefore the PR as it is afaik without any detriments^^

Aha of course :D

pdaniell-nv:

PR #425, opened 2020-08-24
glx: Add GLX_EXT_get_drawable_type

nwnk:
While writing some other client library bits, I discovered a case where I wanted to differentiate based on the type of the GLXDrawable, but that there's no query for that. It was trivial to add, so I added it. This extension formalizes it. I picked EXT because this seems uncontroversial and easy to support, but I'm happy to change it to MESA or XORG or RHT or whatever if there's no other interest.

### DIFF
--- a/extensions/EXT/GLX_EXT_get_drawable_type.txt
+++ b/extensions/EXT/GLX_EXT_get_drawable_type.txt
@@ -1,0 +1,93 @@
+Name
+
+    EXT_get_drawable_type
+
+Name Strings
+
+    GLX_EXT_get_drawable_type
+
+Contributors
+
+    Adam Jackson, Red Hat
+    Kyle Brenneman, NVIDIA
+
+Contacts
+
+    Adam Jackson <ajax@redhat.com>
+
+Status
+
+    Complete.
+
+Version
+
+    Version 3, 24 September 2020
+
+Number
+
+    OpenGL Extension #552
+
+Dependencies
+
+    GLX 1.3 or higher is required.
+
+    This extension is written against the text of the GLX 1.4 specification.
+
+Overview
+
+    This extension adds to the set of drawable attributes that are sent in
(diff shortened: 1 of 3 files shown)


kbrenneman:
This is more of a minor bikeshedding note, but I'd recommend a name like "GLX_EXT_get_drawable_type", since that's the attribute that it adds.

extensions/EXT/GLX_EXT_get_drawable_type.txt
```diff
@@ -0,0 +1,91 @@
...
+    Adam Jackson, Red Hat
+    Kyle Brenneman, NVIDIA
+
+Contacts
+
+    Adam Jackson <ajax@redhat.com>
+
+Status
+
+    Incomplete
```
cubanismo, line 20: Assuming this is actually complete now, LGTM.

cubanismo (changes requested):
I think you need to update externsions/registry.py as well, assign yourself an extension number and include it here instead of XXX.  There may be other housekeeping.  I haven't done a GLX extension since the registry was converted to the new format.  Run through the Readme.md and make sure it all still builds and shows up in the generated web pages.  Other than that, the spec content looks good to me.

pdaniell-nv:
This can be merged once the XXX and "Incomplete" status issues are resolved.

nwnk:
Updated the registry and generated docs. Not sure why this still shows as Changes Requested though.

cubanismo (approved):

cubanismo:
LGTM.  Not sure who to assign to merge?  @pdaniell-nv @oddhack?

pdaniell-nv:
@oddhack, this is approved to merge. Thanks.

oddhack:
This is missing the changes to glx.xml but I'll just deal with that, since it's so simple.

Issue #361, opened 2020-01-27
[xml] enum parameters in <commands> without group attribute


SunSerega:
There are 259 of these:
[commands with enum without group.txt](https://github.com/KhronosGroup/OpenGL-Registry/files/4115325/commands.with.enum.without.group.txt)
(i account parameters with `GLenum` and `GLbitfield` types)

I'm not entirely sure if it's worth fixing, with it's scale. Because it can't really be full automated, someone needs to check every function individually.

But i guess it should be at least documented, that you can't rely on `group` attribute being there...

oddhack:
This is something @Perksey might pick up as part of the improvements they've been working on.

Perksey:
Ah, thanks for the ping. I'll add this to the list 👍

Perksey:
After looking at this, I don't think this particular issue warrants a fix. It's too big of a job and involves changing a lot of historical content which likely is surrounded with lots of unknowns.

This issue may be closed as wontfix.

oddhack:
@SunSerega are you OK doing as @Perksey suggests above?

SunSerega:
Well. i'm okay with not fixing this from commands of deprecated extensions and such.

But from what listed here - some are from relevant extensions. And some are from core.

Maybe it makes sense to break this issue in multiple once, so that core can be fixed with more priority?

And about extension once - i think, for it to be manageable, .xml first need to have a way to mark extension as non-relevant.

Perksey:
Groups are not, and should not, be a requirement for the entirety of the gl.xml and are really just niceties left over from the demise of Silicon Graphics. As such, unless you personally intend to fix this issue, I don’t advise taking this further.

SunSerega:
I'm not saying they are required for C/C++ gl header generation. And so i'm not implying in any way that khronos needs to deal with this. Group are needed only for high lvl languages binding, so, obviously, this issue should be dealt by someone who does this kind of binding.

Do you think that because it's not required for what khronos directly manages - this is not a place to make such issues? Do you know better place?

> unless you personally intend to fix this

I thought i said it already, i'm going to fix everything i post when i have time. I haven't had much of it recently, with all the irl stuff... But i'm not forgetting about these issues.

> It's too big of a job

And i think this is wrong way of thinking. Only result should dictate if work needs to be done, not things like amount of work or how interesting it is.

If it's too big - it needs to be broken up into smaller jobs. This is why i started suggesting things like separating core functions and starting from them.

[...]

PR #427, opened 2020-08-28
[xml] Missing group attributes on enum/bitfield arguments of core functions

SunSerega:
Part of #361

1. Some `glGet*` functions didn't have `group="GetPName"`.

2. `glShaderBinary`:
Specification defines at least one specific value - `GL_SHADER_BINARY_FORMAT_SPIR_V`.
So, i defined `ShaderBinaryFormat` group for it.\
\
Besides that - while looking for info i found quite a few people passing 0 as binary format. Is it some kind of special value "general format"? Should it also be added to `ShaderBinaryFormat` group?

3. `glFenceSync` and `glWaitSync` have `flags` parameter. It's reserved for now, so only 0 can be passed into them.
I added  `EmptyFlags` group, so that this restriction can be easier enforced in higl-level bindings.

I also applied all the same changes to extension functions, that define these functions as `<alias>`.

---

Somewhat off topic but related questions:

- `glSampleMaski` takes `GLbitfield` argument. And, well, it is a bitfield, but it's not what `GLbitfield` is usually used for.
Maybe it would make more sense to change that argument to `GLuint`, so it's not confused with actual enum arguments?

- I found and added few missing `group="sync"` attributes's in extension functions, alias to core functions that have this attribute.
But can someone actually explain why most arguments of type `GLsync` also have group `sync`?
Is it an old thing, from when there wasn't a `GLsync` type? Does it still need to exist?


xml/gl.xml
```diff
@@ -5234,7 +5234,7 @@ typedef unsigned int GLhandleARB;
         <enum value="0" name="GL_FALSE" group="Boolean,VertexShaderWriteMaskEXT,ClampColorModeARB"/>
         <enum value="0" name="GL_NO_ERROR" group="GraphicsResetStatus,ErrorCode"/>
         <enum value="0" name="GL_ZERO" group="TextureSwizzle,StencilOp,BlendingFactor"/>
-        <enum value="0" name="GL_NONE" group="TextureCompareMode,PathColorFormat,CombinerBiasNV,CombinerScaleNV,DrawBufferMode,PixelTexGenMode,ReadBufferMode,ColorBuffer,PathGenMode,PathTransformType,PathFontStyle"/>
+        <enum value="0" name="GL_NONE" group="EmptyFlags,TextureCompareMode,PathColorFormat,CombinerBiasNV,CombinerScaleNV,DrawBufferMode,PixelTexGenMode,ReadBufferMode,ColorBuffer,PathGenMode,PathTransformType,PathFontStyle"/>
```
Perksey, line 5237: I wouldn't call this `EmptyFlags` as it's not future-proof. Perhaps `SyncBehaviorFlags`? (Yes, I know that there's no currently defined flags for the glFenceSync flags parameter, but maybe one day there will be...)

Perksey (changes requested):
Thank you so much for this PR! Most of the groups look good, however...

xml/gl.xml
SunSerega: Well, this only part of core functions, i imagine there would be more of empty enum/bitfield arguments in extensions.

They probably not getting any enum values any time soon, so it makes sense to use 2 common groups in all functions - `EmptyEnum` and `EmptyFlags`, to not clog up xml files. And then change group name if these functions do get enum values.

It shouldn't be any problem to change group name later, right?

Perksey: I really think we should keep GLenum and GLbitfield parameters without associated enumerants untagged, Empty* is a smell and a little pointless. Group changes later down the line are technically breaking if concerning the function signatures.

SunSerega: > Group changes later down the line are technically breaking if concerning the function signatures.

Well, adding missing `group="GetPName"` would also break signatures then...
But i guess changes that break signatures can at least be minimized.

Perksey (approved):

pdaniell-nv:
@oddhack this is approved to merge assuming you're happy with it.

PR #435, opened 2020-10-28
[xml] Missing class=sync

SunSerega:
Small update to #428.

When updating my bindings to use classes i noticed - i missed few functions returning `GLsync` type...
For most other classes return type is `GLuint`, so i wasn't checking functions by return type.


pdaniell-nv:
@oddhack if you're happy with this change it is approved to merge.

PR #436, opened 2020-10-30
Add NV_primitive_shading_rate extension.

nvpbrown:
This extension builds on top of the NV_shading_rate_image extension to
provide OpenGL API support for using a per-primitive shading rate
value to control the computation of the rate used to process each
fragment.

--

I think I have prepared this pull request successfully.  I did build headers and other materials, but my understanding is that I'm supposed to leave this out of a pull request and that those generated files will be rebuilt at the time the extensions are merged and/or published.  Let me know if there are any issues with this submission.

Thanks,
Pat

@pdaniell-nv

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -11586,7 +11586,8 @@ typedef unsigned int GLhandleARB;
         <enum value="0x95AE" name="GL_SHADING_RATE_SAMPLE_ORDER_DEFAULT_NV"/>
         <enum value="0x95AF" name="GL_SHADING_RATE_SAMPLE_ORDER_PIXEL_MAJOR_NV"/>
         <enum value="0x95B0" name="GL_SHADING_RATE_SAMPLE_ORDER_SAMPLE_MAJOR_NV"/>
-        <unused start="0x95B1" end="0x95B2" vendor="NV"/>
+        <enum value="0x95B1" name="GL_SHADING_RATE_IMAGE_PER_PRIMITIVE_NV"/>
+        <enum value="0x95B2" name="GL_SHADING_RATE_IMAGE_PALETTE_COUNT_NV"/>
         <enum value="0x95B3" name="GL_SEMAPHORE_TYPE_NV" group="SemaphoreParameterName"/>
         <enum value="0x95B4" name="GL_SEMAPHORE_TYPE_BINARY_NV" group="SemaphoreParameterName"/>
         <enum value="0x95B5" name="GL_SEMAPHORE_TYPE_TIMELINE_NV" group="SemaphoreParameterName"/>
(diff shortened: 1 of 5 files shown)


xml/gl.xml
```diff
@@ -11586,7 +11586,8 @@ typedef unsigned int GLhandleARB;
         <enum value="0x95AE" name="GL_SHADING_RATE_SAMPLE_ORDER_DEFAULT_NV"/>
         <enum value="0x95AF" name="GL_SHADING_RATE_SAMPLE_ORDER_PIXEL_MAJOR_NV"/>
         <enum value="0x95B0" name="GL_SHADING_RATE_SAMPLE_ORDER_SAMPLE_MAJOR_NV"/>
-        <unused start="0x95B1" end="0x95B2" vendor="NV"/>
+        <enum value="0x95B1" name="GL_SHADING_RATE_IMAGE_PER_PRIMITIVE_NV"/>
+        <enum value="0x95B2" name="GL_SHADING_RATE_IMAGE_PALETTE_COUNT_NV"/>
```
Perksey, line 11590:
```suggestion
        <enum value="0x95B1" name="GL_SHADING_RATE_IMAGE_PER_PRIMITIVE_NV" group="EnableCap,GetPName"/>
        <enum value="0x95B2" name="GL_SHADING_RATE_IMAGE_PALETTE_COUNT_NV" group="GetPName"/>
```

Perksey (changes requested):

Perksey:
Also, yeah header regeneration is generally done by Jon after merge mainly so pull requests don't conflict with one another.

xml/gl.xml
nvpbrown: I don't really understand the function of all this, but it sounds like it's a feature that might be used by some third-party tools and not the direct Khronos scripts.  Based on a review of other similar logic, this proposal seems reasonable for the most part.

BUT just to check, if we make this edit, do we also need to add the enums to the corresponding sections referenced by the "group" attribute?  It sounds like maybe not, based on the "Deprecated" comment here?
```
        <group name="EnableCap" comment="Deprecated, use the group attributes instead.">
            <enum name="GL_ALPHA_TEST"/>
            <enum name="GL_ASYNC_DRAW_PIXELS_SGIX"/>
```

nvpbrown: I tried to use the "Commit suggestion" button here to merge the proposed changes, but I kept getting "This diff has recently been updated.  Refresh and try again."  It never worked, so I pushed an identical commit myself.

oddhack: @nvpbrown you are correct, that stuff is solely for downstream consumers. It should not hold up publication if that's important to you for business reasons.

oddhack:
@nvpbrown @pdaniell-nv this is fine. If Piers wants to put the Approved milestone on it I'll merge it & generate headers.

pdaniell-nv:
> @nvpbrown @pdaniell-nv this is fine. If Piers wants to put the Approved milestone on it I'll merge it & generate headers.

@oddhack, thanks. Approved to merge.

PR #438, opened 2020-11-08
Update len attributes

frederikja163:
```c
glGetnUniformdv
glGetnUniformuiv
glGetPerfMonitorCounterData
glGetnMapdv
glGetnPixelMapfv
```
All have a parameter with a length attribute for a parameter which is not of type `void*` or `const void*`. The length is specified in byte count which makes its length attribute incorrect as its written like if it was an element count.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -20745,7 +20745,7 @@ typedef unsigned int GLhandleARB;
             <param group="MapTarget"><ptype>GLenum</ptype> <name>target</name></param>
             <param group="MapQuery"><ptype>GLenum</ptype> <name>query</name></param>
             <param><ptype>GLsizei</ptype> <name>bufSize</name></param>
-            <param len="COMPSIZE(bufSize)"><ptype>GLdouble</ptype> *<name>v</name></param>
+            <param len="bufSize / 8"><ptype>GLdouble</ptype> *<name>v</name></param>
         </command>
         <command>
             <proto>void <name>glGetnMapfv</name></proto>
@@ -20803,7 +20803,7 @@ typedef unsigned int GLhandleARB;
             <proto>void <name>glGetnPixelMapfvARB</name></proto>
             <param group="PixelMap"><ptype>GLenum</ptype> <name>map</name></param>
             <param><ptype>GLsizei</ptype> <name>bufSize</name></param>
-            <param len="COMPSIZE(bufSize)"><ptype>GLfloat</ptype> *<name>values</name></param>
+            <param len="bufSize / 4"><ptype>GLfloat</ptype> *<name>values</name></param>
         </command>
         <command>
             <proto>void <name>glGetnPixelMapuiv</name></proto>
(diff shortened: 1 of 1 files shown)


pdaniell-nv:
@oddhack this change is approved to merge if you're happy with it.

xml/gl.xml
```diff
@@ -19132,7 +19132,7 @@ typedef unsigned int GLhandleARB;
             <param><ptype>GLuint</ptype> <name>monitor</name></param>
             <param><ptype>GLenum</ptype> <name>pname</name></param>
             <param><ptype>GLsizei</ptype> <name>dataSize</name></param>
-            <param len="dataSize"><ptype>GLuint</ptype> *<name>data</name></param>
+            <param len="COMPSIZE(dataSize)"><ptype>GLuint</ptype> *<name>data</name></param>
```
Perksey, line 19135: COMPSIZE isn't really needed for single-parameter counts, mainly because there's nothing to compute. This is just an inconsistency in some places that the best solution for fixing would be removing unnecessary COMPSIZE from single parameters, rather than adding them elsewhere.

frederikja163: the size of the value is used aswell tho

Perksey: `len="dataSize"` implicitly takes into account the size of the parameter type, as seen elsewhere within the spec. `len` refers to the number of `ptype`s in `ptype*`.

The spec is horribly inconsistent about this, however, and that's probably because COMPSIZE isn't even a documented construct in any of the OpenGL specification specifications such as the schema and readme.tex.

frederikja163: i am aware of this, but this will at the very least make it more consistent.

oddhack:
As noted by @Perksey COMPSIZE is essentially meaningless, a historical curiosity. I have exactly zero investment in whether it stays or goes but since @Perksey raised concerns after @pdaniell-nv marked this Approved, I'm going to hold off on merging it until the concerned parties hash it out between themselves.

oddhack:
Un-Approving until there's no dissension.

xml/gl.xml
Perksey: To be honest removing COMPSIZE would be the best option, however we can't feasibly do that in case older generators using the OpenGL spec use it.

Instead we should look to phase it out by not adding it anywhere else. In addition, we should discuss whether we can use multi-parameter counts without COMPSIZE, but that's probably best suited for its own discussion.

COMPSIZE isn't itself intrinsically meaningful and is a remnant of the old gl.spec format. It shouldn't be used for generators using gl.xml due to its undefined functionality, and we should avoid adding it throughout the spec for this reason as well.

frederikja163: It is not meaningless, and if you cant use it for your projects you can just ignore all compsizes, but we are currently using compsize internally in opentk with our new binder.

frederikja163: Compsize tells us what parameters are used for the length of an array, and it tells us that it is not just their value we should use as the size but rather that the size is computed from said values

SunSerega: Btw, `len` attribute actually supports any expressions (though it's rarely used):
```
        <command>
            <proto>void <name>glFogFuncSGIS</name></proto>
            <param><ptype>GLsizei</ptype> <name>n</name></param>
            <param len="n*2">const <ptype>GLfloat</ptype> *<name>points</name></param>
            <glx type="render" opcode="2067"/>
        </command>
```
> `len` - parameter length, either an integer specifying the number of elements of
the parameter `<ptype>`, **or a complex string expression with poorly defined
syntax**, usually representing a length that is computed as a combination of other
command parameter values, and possibly current GL state as well.

frederikja163: I am aware that it supports this, and maybe we should use this. But in that case i think we really should add this to a lot of other methods aswell.

oddhack: COMPSIZE isn't "intrinsically" meaningful because what it really meant in the original SGI spec files was "this needs to have handwritten GLX wire protocol encode / decode, too complex to automatically generate".

Perksey: Fair enough, in that case I don't think we should use it in places where GLX doesn't get involved. Though, it should probably still be used in multi-parameter counts because a) most of the time we can't calculate that in a defined way and b) existing parsers may break if we get multiple parameters not enclosed within a compsize.

frederikja163: Would people be happy with a multiplier instead then? Like n*4 for float for example

SunSerega: I think it's at least better then `COMPSIZE`... Except it should be divided, not multiplied. And it's probably better to use `sizeof`, instead of magic numbers.

Perksey: n*4 implies a mathematical calculation (and has been used in that way elsewhere in the spec. Could come up with an alternative, but it will also likely break other parsers. I think it's a case of "we're stuck with this smell but let's not make it any worse"

Perksey: Oh right, just got what you're on about. It won't be clear to readers of the spec what the *4 means. I think it's just best to treat all lengths as implicitly multiplied by the size of the element type.

SunSerega: Meanwhile i thought about something like this:
```
len = "dataSize / sizeof(GLuint)"
```

Perksey: I'm primarily worried about other parsers seeing as there's no easy way we can have a slow migration over to any new syntax we introduce. There's not really anything we can do other than not make the problem any worse

frederikja163: SunSerega it would still be multiply and not divide. Also, i dont see the issue with multiplying by a constant as the types have a constant size, plus it wouldnt be introducing new syntax like perksey touched on

Perksey: In a number of places the number in the len attribute is actually the number of the type (len is used **very** extensively this way throughout the spec, just open up the xml and ctrl+f for `len` and you'll see it used like this everywhere)

frederikja163:
This has now been updated to use divide rather than compsize

Perksey (approved):
LGTM

Perksey:
Eliminating COMPSIZE entirely where possible in favour of mathematical expressions like this could be a cool idea actually but may take some time

PR #439, opened 2020-11-08
glFlushVertexArrayRangeNV

frederikja163:
The NV version of this method was missing all contents for some reason. I modified it to look like the apple command, however im unsure of wether or not this is actually the correct signature

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -16699,6 +16699,8 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glFlushVertexArrayRangeNV</name></proto>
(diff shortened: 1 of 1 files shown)


pdaniell-nv:
The glFlushVertexArrayRangeNV function is different to the glFlushVertexArrayRangeAPPLE function. The NV one takes no parameters, so the current XML is correct. See https://www.khronos.org/registry/OpenGL/extensions/NV/NV_vertex_array_range.txt.

frederikja163:
Ok, i just doublechecked and i found my information at some incorrect third party sites, sorry for the inconvinience

PR #440, opened 2020-11-08
Fix param len attribute on multiple functions to match their ARB counterparts

NogginBops:
There where a few functions that had `<param>`s there where missing a `len` attribute while the functions ARB counterpart had them.
This PR fixes these inconsistencies.

`glGetnMapdv` and `glGetnPixelMapfv` and their ARB counterparts where also modified to use `COMPSIZE` as their length depends on the pointer element size as well.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -20832,7 +20832,7 @@ typedef unsigned int GLhandleARB;
         <command>
             <proto>void <name>glGetnPolygonStipple</name></proto>
             <param><ptype>GLsizei</ptype> <name>bufSize</name></param>
-            <param><ptype>GLubyte</ptype> *<name>pattern</name></param>
+            <param len="bufSize"><ptype>GLubyte</ptype> *<name>pattern</name></param>
         </command>
         <command>
             <proto>void <name>glGetnPolygonStippleARB</name></proto>
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -20845,10 +20845,10 @@ typedef unsigned int GLhandleARB;
             <param group="PixelFormat"><ptype>GLenum</ptype> <name>format</name></param>
             <param group="PixelType"><ptype>GLenum</ptype> <name>type</name></param>
             <param><ptype>GLsizei</ptype> <name>rowBufSize</name></param>
-            <param>void *<name>row</name></param>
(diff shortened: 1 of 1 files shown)


frederikja163:
The changes for the two methods: glGetnMapdv, glGetnPixelMapfv were also reflected and implented in my PR #438

pdaniell-nv:
@oddhack this change is approved to merge if you're happy with it.

xml/gl.xml
```diff
@@ -20738,14 +20738,14 @@ typedef unsigned int GLhandleARB;
             <param group="MapTarget"><ptype>GLenum</ptype> <name>target</name></param>
             <param group="MapQuery"><ptype>GLenum</ptype> <name>query</name></param>
             <param><ptype>GLsizei</ptype> <name>bufSize</name></param>
-            <param><ptype>GLdouble</ptype> *<name>v</name></param>
+            <param len="COMPSIZE(bufSize)"><ptype>GLdouble</ptype> *<name>v</name></param>
```
Perksey, line 20741: Same comment as in #438

NogginBops: This is intentional and will not result in a merge conflict.

Perksey: From #438:

> COMPSIZE isn't really needed for single-parameter counts, mainly because there's nothing to compute. This is just an inconsistency in some places that the best solution for fixing would be removing unnecessary COMPSIZE from single parameters, rather than adding them elsewhere.
> 
> COMPSIZE is only needed where multiple parameters are used as part of the length calculation

NogginBops:
With some further looking by @frederikja163, we found around 180 other functions similar to this where the ARB version has a `len` attribute while the non-ARB version doesn't.

xml/gl.xml
oddhack: I don't think that's true everywhere although it may be mostly true. I believe there's a couple of places where a count is given in units different than the underlying type. What COMPSIZE indicates is that additional logic is needed to compute the actual size from the parameters. That said we use COMPSIZE for exactly nothing; it is a 30 year holdover from SGI's original .spec files and the actual logic required always had to be externally defined. So I don't really care whether it stays or goes.

PR #441, opened 2020-11-11
Update maxdrawcount parameter type in ARB_indirect_parameters to be consistent

pdaniell-nv:
Fixes issue https://github.com/KhronosGroup/OpenGL-API/issues/77.

The "maxdrawcount" parameter type wasn't self-consistent within the ARB_indirect_parameters extension specification. It should be the same "sizei" type everywhere in the spec.

### DIFF
--- a/extensions/ARB/ARB_indirect_parameters.txt
+++ b/extensions/ARB/ARB_indirect_parameters.txt
@@ -36,8 +36,8 @@ Status
 
 Version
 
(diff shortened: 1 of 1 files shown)


oddhack:
Looks fine, consistent with API spec.

pdaniell-nv:
@oddhack this is approved to merge. We can close https://github.com/KhronosGroup/OpenGL-API/issues/77 once merged.

PR #445, opened 2020-11-19
Fix code sample in EXT_disjoint_timer_query

xndcn:
This fixes incomplete code in example (2) and wrong parameter order of `glQueryCounterEXT` in example (3)


pdaniell-nv:
@oddhack if you're happy with this it's approved.

PR #448, opened 2020-11-21
GL_SHADER_BINARY_FORMATS included in shader binary format

frederikja163:
`glShaderBinary` had no supported enums from v4.1 to 4.5 as `GL_SHADER_BINARY_FORMAT_SPIR_V` was only added in 4.6
According to the documentation however it seems like GL_SHADER_BINARY_FORMATS is also a supported enum. Unless i misunderstood something, in which case i will hapily change this.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -9324,7 +9324,7 @@ typedef unsigned int GLhandleARB;
         <enum value="0x8C07" name="GL_FACTOR_ALPHA_MODULATE_IMG"/>
         <enum value="0x8C08" name="GL_FRAGMENT_ALPHA_MODULATE_IMG"/>
         <enum value="0x8C09" name="GL_ADD_BLEND_IMG"/>
-        <enum value="0x8C0A" name="GL_SGX_BINARY_IMG"/>
+        <enum value="0x8C0A" name="GL_SGX_BINARY_IMG" group="ShaderBinaryFormat"/>
             <unused start="0x8C0B" end="0x8C0F" vendor="IMG"/>
     </enums>
 
@@ -10319,7 +10319,7 @@ typedef unsigned int GLhandleARB;
     </enums>
 
     <enums namespace="GL" start="0x8F60" end="0x8F6F" vendor="ARM" comment="For Remi Pedersen, Khronos bug 3745">
-        <enum value="0x8F60" name="GL_MALI_SHADER_BINARY_ARM"/>
+        <enum value="0x8F60" name="GL_MALI_SHADER_BINARY_ARM" group="ShaderBinaryFormat"/>
         <enum value="0x8F61" name="GL_MALI_PROGRAM_BINARY_ARM"/>
             <unused start="0x8F62" vendor="ARM"/>
         <enum value="0x8F63" name="GL_MAX_SHADER_PIXEL_LOCAL_STORAGE_FAST_SIZE_EXT"/>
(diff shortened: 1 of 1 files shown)


oddhack:
SHADER_BINARY_FORMATS isn't a shader format you can pass to ShaderBinary; it's the token you query with GetIntegerv to return the list of implementation-supported formats, which will generally just be [ SHADER_BINARY_FORMAT_SPIR_V ]. So it shouldn't be tagged that way.

xml/gl.xml
```diff
@@ -9982,7 +9982,7 @@ typedef unsigned int GLhandleARB;
         <enum value="0x8DF5" name="GL_HIGH_INT" group="PrecisionType"/>
         <enum value="0x8DF6" name="GL_UNSIGNED_INT_10_10_10_2_OES"/>
         <enum value="0x8DF7" name="GL_INT_10_10_10_2_OES"/>
-        <enum value="0x8DF8" name="GL_SHADER_BINARY_FORMATS"/>
+        <enum value="0x8DF8" name="GL_SHADER_BINARY_FORMATS" group="ShaderBinaryFormat"/>
```
Perksey, line 9985:
```suggestion
        <enum value="0x8DF8" name="GL_SHADER_BINARY_FORMATS" group="GetPName"/>
```

Perksey (changes requested):

frederikja163:
> SHADER_BINARY_FORMATS isn't a shader format you can pass to ShaderBinary; it's the token you query with GetIntegerv to return the list of implementation-supported formats, which will generally just be [ SHADER_BINARY_FORMAT_SPIR_V ]. So it shouldn't be tagged that way.

Are there any other shader binary formats? Cause the problem with SPIR-V is that it is only available on core from 4.5 and forwards whereas the ARBes2 compat extension that introduced the ShaderBinaryFormat enum was included in core from 4.1 and onwards. So im wondering what enums you would use for 4.2 without the SPIR-V binary being available

Perksey:
ARB_es2_compatibility didn't specify any formats in its original specification.

dgkoch:
>  So im wondering what enums you would use for 4.2 without the SPIR-V binary being available

You can't unless there are vendor extensions to add them. 
The ARB_es2_compatibility spec even says: 

    OpenGL defines no specific binary formats, but does provide a mechanism to obtain
    token values for such formats provided by extensions.

> Are there any other shader binary formats?

Taking a quick scan through the registry in ES land there is: 
- IMG_shader_binary
- ARM_mali_shader_binary
- VIV_shader_binary
- DMP_shader_binary
- FJ_shader_binary_GCCSO
- NV_platform_binary

that all add ShaderBinary formats. I don't know of any that were specifically added to GL, but that doesn't mean there weren't some used that just never got registered. The main goal of adding this to GL was API convergence.

xml/gl.xml
dgkoch: The original change is definitely not correct.

pdaniell-nv:
@frederikja163 are you planning to update this PR with the suggested change?

frederikja163:
Yes, i just havent had that much time recently, but i have time again now. So i will take a look at it tomorrow

frederikja163:
Couldnt find the NV platform binary, but the rest should be added here. Still wondering tho what you should call the method with on a opengl 4.2 context without any of those extensions supported

SunSerega:
And I, for a long time, though you can only obtain this format from `glGetProgramBinary` and only pass it to `glShaderBinary` and `glProgramBinary`.

oddhack:
> Couldnt find the NV platform binary, but the rest should be added here. Still wondering tho what you should call the method with on a opengl 4.2 context without any of those extensions supported

You shouldn't call it, because there are no supported formats.

The alternative would have been to have a separate query in every extension introducing a binary format (or have each of them implement the same query with a different supported token).

Perksey (approved):
LGTM

pdaniell-nv:

PR #449, opened 2020-12-01
Add QCOM_frame_extrapolation

tate-hornbeck:


SunSerega:
It's not a problem, but you don't need to add `group="Texture"` to new `<param>` elements because the newer `class` attribute does the same, but more consistently.

xml/gl.xml
```diff
@@ -28915,6 +28915,13 @@ typedef unsigned int GLhandleARB;
             <param class="texture" group="Texture"><ptype>GLuint</ptype> <name>output</name></param>
             <param class="texture" group="Texture"><ptype>GLuint</ptype> <name>mask</name></param>
         </command>
+        <command>
+            <proto>void <name>glExtrapolateTex2DQCOM</name></proto>
+            <param class="texture" group="Texture"><ptype>GLuint</ptype> <name>src1</name></param>
+            <param class="texture" group="Texture"><ptype>GLuint</ptype> <name>src2</name></param>
+            <param class="texture" group="Texture"><ptype>GLuint</ptype> <name>output</name></param>
```
SunSerega, line 28922:
```suggestion
            <param class="texture"><ptype>GLuint</ptype> <name>src1</name></param>
            <param class="texture"><ptype>GLuint</ptype> <name>src2</name></param>
            <param class="texture"><ptype>GLuint</ptype> <name>output</name></param>
```

Perksey: The class attribute is pretty new though, so it's probably best to keep the group as well to keep consistent with other functions and keep compatibility with older binders which may make some meaningful use out of GLuint groups (ANGLE springs to mind)

SunSerega: Usage of this kind of `group` attribute, in this case, is inconsistent either way: There are currently 176 entries of `class="texture"` and only 74 of `group="Texture"` in `gl.xml`. This was one of the main reasons to start making `class` attribute.

Old bindings are a good point though. In the future I'm going to ask for the removal of `group` attributes that dup `class` attributes, just like with the old `<groups>` block.  So I thought it's better to not even add this kind of `group` attributes in the meanwhile. But if there are still bindings that could benefit from them - maybe they should still be added.

Perksey: For that you'll likely have to communicate with the bindings authors and, once they are prepared for the change, add a resonable window between an obselesence notice being posted and removing them.

Perksey: Anyways I stress again, there's no problem at all going either way but seeing as the suggested change can be applied with the click of a button, feel free to add these attributes.

Perksey (approved):
LGTM!

pdaniell-nv:
@oddhack this is approved to merge if you're happy with it.

PR #451, opened 2020-12-11
Add GL_ARB_ES1_compatibility

caramelli:
I am using OpenGL ES 1.1 code with an OpenGL implementation (GL headers and libGL library).
Everything is OK except the declaration of 4 prototypes which are missing for compatibility.
So I just define GL_ARB_ES1_compatibility to add them (similar to GL_ARB_ES2_compatibility).
But that may not be the right way to do it.

Nicolas Caramelli


pdaniell-nv:
Doing an extension requires more than just adding the functions to the XML, it also needs to be documented (see https://www.khronos.org/registry/OpenGL/index_gl.php). You would probably make quicker progress making this a vendor or EXT extension rather than an ARB extension. ARB extensions require ratification from the Khronos board, as well as 3+ implementations and CTS.

Who are you hoping will implement this? NVIDIA have something similar called GL_NV_ES1_1_compatibility, which was never published. That extension seems to expose a lot more functions than the 4 you have exposed here.

``` c++
GLAPI void GLAPIENTRY glAlphaFuncx (GLenum func, GLclampx ref);
GLAPI void GLAPIENTRY glClearColorx (GLclampx red, GLclampx green, GLclampx blue, GLclampx alpha);
GLAPI void GLAPIENTRY glClearDepthx (GLclampx depth);
GLAPI void GLAPIENTRY glClipPlanef (GLenum plane, const GLfloat *equation);
GLAPI void GLAPIENTRY glClipPlanex (GLenum plane, const GLfixed *equation);
GLAPI void GLAPIENTRY glColor3x (GLfixed red, GLfixed green, GLfixed blue);
GLAPI void GLAPIENTRY glColor4x (GLfixed red, GLfixed green, GLfixed blue, GLfixed alpha);
GLAPI void GLAPIENTRY glDepthRangex (GLclampx zNear, GLclampx zFar);
GLAPI void GLAPIENTRY glFogx (GLenum pname, GLfixed param);
GLAPI void GLAPIENTRY glFogxv (GLenum pname, const GLfixed *params);
GLAPI void GLAPIENTRY glFrustumf (GLfloat left, GLfloat right, GLfloat bottom, GLfloat top, GLfloat zNear, GLfloat zFar);
GLAPI void GLAPIENTRY glFrustumx (GLfixed left, GLfixed right, GLfixed bottom, GLfixed top, GLfixed zNear, GLfixed zFar);
GLAPI void GLAPIENTRY glGetClipPlanef (GLenum plane, GLfloat *eqn);
GLAPI void GLAPIENTRY glGetClipPlanex (GLenum plane, GLfixed *eqn);
GLAPI void GLAPIENTRY glGetFixedv (GLenum pname, GLfixed *params);
GLAPI void GLAPIENTRY glGetLightxv (GLenum light, GLenum pname, GLfixed *params);
GLAPI void GLAPIENTRY glGetMaterialxv (GLenum face, GLenum pname, GLfixed *params);
GLAPI void GLAPIENTRY glGetTexEnvxv (GLenum env, GLenum pname, GLfixed *params);
GLAPI void GLAPIENTRY glGetTexParameterxv (GLenum target, GLenum pname, GLfixed *params);
GLAPI void GLAPIENTRY glLightModelx (GLenum pname, GLfixed param);
GLAPI void GLAPIENTRY glLightModelxv (GLenum pname, const GLfixed *params);
GLAPI void GLAPIENTRY glLightx (GLenum light, GLenum pname, GLfixed param);
GLAPI void GLAPIENTRY glLightxv (GLenum light, GLenum pname, const GLfixed *params);
GLAPI void GLAPIENTRY glLineWidthx (GLfixed width);
GLAPI void GLAPIENTRY glLoadMatrixx (const GLfixed *m);
GLAPI void GLAPIENTRY glMaterialx (GLenum face, GLenum pname, GLfixed param);
GLAPI void GLAPIENTRY glMaterialxv (GLenum face, GLenum pname, const GLfixed *params);
GLAPI void GLAPIENTRY glMultiTexCoord1x (GLenum target, GLfixed s);
GLAPI void GLAPIENTRY glMultiTexCoord1xv (GLenum target, const GLfixed *v);
GLAPI void GLAPIENTRY glMultiTexCoord2x (GLenum target, GLfixed s, GLfixed t);
GLAPI void GLAPIENTRY glMultiTexCoord2xv (GLenum target, const GLfixed *v);
GLAPI void GLAPIENTRY glMultiTexCoord3x (GLenum target, GLfixed s, GLfixed t, GLfixed r);
GLAPI void GLAPIENTRY glMultiTexCoord3xv (GLenum target, const GLfixed *v);
GLAPI void GLAPIENTRY glMultiTexCoord4x (GLenum target, GLfixed s, GLfixed t, GLfixed r, GLfixed q);
GLAPI void GLAPIENTRY glMultiTexCoord4xv (GLenum target, const GLfixed *v);
GLAPI void GLAPIENTRY glMultMatrixx (const GLfixed *m);
GLAPI void GLAPIENTRY glNormal3x (GLfixed nx, GLfixed ny, GLfixed nz);
GLAPI void GLAPIENTRY glOrthof (GLfloat left, GLfloat right, GLfloat bottom, GLfloat top, GLfloat zNear, GLfloat zFar);
GLAPI void GLAPIENTRY glOrthox (GLfixed left, GLfixed right, GLfixed bottom, GLfixed top, GLfixed zNear, GLfixed zFar);
GLAPI void GLAPIENTRY glPointParameterx (GLenum pname, GLfixed param);
GLAPI void GLAPIENTRY glPointParameterxv (GLenum pname, const GLfixed *params);
GLAPI void GLAPIENTRY glPointSizex (GLfixed size);
GLAPI void GLAPIENTRY glPolygonOffsetx (GLfixed factor, GLfixed units);
GLAPI void GLAPIENTRY glRotatex (GLfixed angle, GLfixed x, GLfixed y, GLfixed z);
GLAPI void GLAPIENTRY glSampleCoveragex (GLclampx value, GLboolean invert);
GLAPI void GLAPIENTRY glScalex (GLfixed x, GLfixed y, GLfixed z);
GLAPI void GLAPIENTRY glTexEnvx (GLenum target, GLenum pname, GLfixed param);
GLAPI void GLAPIENTRY glTexEnvxv (GLenum target, GLenum pname, const GLfixed *params);
GLAPI void GLAPIENTRY glTexParameterx (GLenum target, GLenum pname, GLfixed param);
GLAPI void GLAPIENTRY glTexParameterxv (GLenum target, GLenum pname, const GLfixed *params);
GLAPI void GLAPIENTRY glTranslatex (GLfixed x, GLfixed y, GLfixed z);
```

Perksey:
+1 The reason these functions aren't present is because no implementation exposes them. Generally it's the people writing the drivers that update the spec with the cool new features they've decided to put into their driver. There's not much we can do not being driver developers I'm afraid.

caramelli:
I do not consider the fixed-point support defined for the common lite profile.
For compatibility in this case, if I'm not wrong, there are only these 4 functions for which you just need to add the prototype.
With Mesa for example, these functions are already implemented in the libGL library, only the prototype is missing.
As suggested, I renamed the proposed change using an EXT extension.
I need to write an extensions/EXT/EXT_ES1_compatibility.txt file, is it correct?

pdaniell-nv:
Sure, if you don't need the fixed-point functions then the 4 you proposed would be enough. And yes you would need to write up a GL_EXT_ES1_compatibility extension.

Perksey:
I'm happy to write up a textual specification if you can get one of the Mesa devs in here so that they know what's going on such that they can support the extension from their end.

nwnk:
Enh. I'm not super thrilled with pretending ES1 is still a thing, but Mesa builds the thunk functions unconditionally so I suppose it's free to advertise. But if we were going to do that I'd prefer we just document the NV extension (where by "we" I guess I mean "NVIDIA"). Practically speaking if Mesa implemented an EXT-named version we'd probably go ahead and expose the NV string as well for compatibility, since if there _is_ any code out there wanting ES1 compat then that's what it's already checking for, so having an EXT string seems a bit pointless.

Perksey:
The NVIDIA extension is significantly larger, so it would may require some work. But if all the functions Piers listed previously are built unconditionally as you said then yeah we could probably go down this route, or maybe promote the NVIDIA extension to EXT as I doubt that NVIDIA will have a lot of cycles going for stuff like this.

nwnk:
> The NVIDIA extension is significantly larger, so it would may require some work. But if all the functions Piers listed previously are built unconditionally as you said then yeah we could probably go down this route, or maybe promote the NVIDIA extension to EXT as I doubt that NVIDIA will have a lot of cycles going for stuff like this.

It's not any larger in the sense that Mesa already implements ES1, all the extra ES1 entrypoints [are thunks into the big-GL versions](https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/src/mesa/main/es1_conversion.c), and we build that code into the driver unconditionally. So adding them to big-GL is really just allowing `glXGetProcAddress` to return them, which we codegen from the list of supported extensions and their entrypoints. I guess there might be some cases that ES1 allows and big-GL1 doesn't, which might require deleting some am-i-desktop-or-not conditionals? That feels unlikely, but I confess I haven't tried to read the ES1 spec in years. 

Also I'm somewhat alarmed to see NVIDIA's driver exposes `GL_NV_ES1_1_compatibility` from core contexts, really not sure what they think `glLoadMatrixx` does in a context that doesn't have a matrix stack. I'm happy to consider that NVIDIA's bug though ;)

"Promoting" the extension to EXT is exactly what I'm not interested in though. If there's any code in the world that wants this, then it already exists and is looking for the NVIDIA string. Adding an EXT version just means they'd have to update their code to check for two things, and any app that didn't do that would fail with an impl that only exposed the EXT string, so we'd end up exposing both _anyway_, so why bother. This isn't unusual, we support plenty of "NV" extensions and at least a few from various vendors that don't exist in the registry IIRC. If all we're really trying to do is make this exist in Mesa then there's no reason to involve the registry, other than it'd be polite for NVIDIA to document it.

Perksey:
I see. In which case @caramelli an issue should be opened for NVIDIA to document this extension. You can use the header snippet that Piers posted to wield these functions, and will be able to use the old NVIDIA string on some NVIDIA devices and Mesa drivers once exposed.

I don't think the EXT extension will be taken further if this is the Mesa's stance, the only other vendor to implement this extension.

caramelli:
Mesa is useful for testing this OpenGL ES 1.1 compatibility, but it goes beyond that. Looking at the links between the different OpenGL specifications, it seems to me that it makes sense to have this compatibility (just 4 missing functions if we don't consider the fixed-point support defined for the common lite profile).

Bringing this consistency has nothing to do with NVIDIA, this unknown compatibility name GL_NV_ES1_1_compatibility seems really strange to me for this kind of compatibility.

Just to be sure that I understood: you suggest to use this name "GL_NV_ES1_1_compatibility" absent from the OpenGL-Registry and to submit a merge request of this "GL_NV_ES1_1_compatibility" extension to Mesa.
It's a limited "compatibility" that doesn't really fit in with the consistency I'm trying to propose.

nwnk:
> Just to be sure that I understood: you suggest to use this name "GL_NV_ES1_1_compatibility" absent from the OpenGL-Registry and to submit a merge request of this "GL_NV_ES1_1_compatibility" extension to Mesa.
> It's a limited "compatibility" that doesn't really fit in with the consistency I'm trying to propose.

What I'm proposing is that the Mesa MR add _everything_ that @pdaniell-nv says is added by `GL_NV_ES1_1_compatibility`, not just the subset you've started with. And then that we just advertise that string, and hopefully NVIDIA gets around to registering it (or, for that matter, someone types up the extension spec for them and they just sign off on it).

I don't really understand why you think that's a "limited compatibility", or what consistency you're trying to propose. To me, what'd be consistent is if everyone exposes the same extension string for the same functionality. And since one vendor's already defined what the extension string is for that functionality, I don't see why we don't just all use that. It's unfortunate that it's not in the registry yet, sure, but that can change, and both NVIDIA and Mesa have a long history of shipping unregistered extensions, so...

Perksey:
Using the existing NV one is more conservative and it allows code to just work that already uses this internal extension (NVIDIA partners?) which is present on many devices today.

Fully in agreement with @nwnk here, but we may require some co-ordination from NVIDIA in registering this extension. @pdaniell-nv I could draft up a spec and perhaps someone from NV could look at it if this is acceptable to you? Of course, ideally NVIDIA should be writing this spec in its entirety, but again I'd imagine you don't have many cycles going for stuff like this.

pdaniell-nv:
Sure if you are willing to knock up a spec I can review and approve. Thanks.

Perksey:
Awesome, can you advise on what tokens this extension has?

pdaniell-nv:
There are no additional tokens beyond what's already in OpenGL compatibility profile, just those function entry points are new. For others that want to implement this extension, once implemented it should pass ES11 CTS.

pdaniell-nv:
In terms of the actual spec content, there are a bunch of places that need to handle GL_FIXED, mostly related to vertex attribute types. I don't see anything else beyond that that needs to be supported at a quick glance.

Perksey:
Affirmative, thanks. Will stitch something together. 

Should the contact for this extension be a NVIDIA contact or should I put my name in there?

caramelli:
Thanks @nwnk for these clarifications! OK, it's not a "limited compatibility" if registered.

Adding everything defined by GL_NV_ES1_1_compatibility is much more important than I had initially imagined!

I never played with the fixed-point support defined for the common lite profile. 
In practice, I'm using the API exposed by GL headers and if I have an OpenGL ES 1.1 target, the purpose is just to avoid replacing glFrustum with glFrustumf for example.
Because I can already use glLightfv, glTexParameteri, glRotatef, glColor4f, ... and all other OpenGL ES 1.1 functions available in the current OpenGL specification, I'm not trying to use fixed-point functions that are not defined. This is why the original change is simply limited to 4 functions!

I have no problem with anything defined by GL_NV_ES1_1_compatibility, but as just explained, the approach is slightly different.
Maybe it's just a name, I understand the conservative argument but "NV" in the name will still be really weird to me ...

@Ultz, is there something I need to do?

Perksey:
> is there something I need to do?

It should be fine for now. I'm going to stitch together a technical, textual specification and update the XML for these functions. Once approved by NVIDIA, this extension will be present in the headers.

pdaniell-nv:
> Should the contact for this extension be a NVIDIA contact or should I put my name in there?

You should probably to mine, unless you want to risk taking responsibility for something that's not totally under your control. You should put your name in the revision history for revision 1.

nwnk:
@caramelli yeah, it's a bit different from what you were proposing, but since Mesa _does_ already have the code for the fixed-point support we may as well expose all of what the NV extension does just to be completely compatible. Thanks for being flexible, I look forward to working with you through the Mesa merge process!

ianromanick:
> I'm not super thrilled with pretending ES1 is still a thing, ...

+1

@nwnk - Are you sure these function should just work on desktop GL?  I _thought_ we would return `GL_INVALID_OPERATION` for functions that don't exist in a particular profile.  ES1 / ES2 / desktop GL have a ton of subtle differences in error handling, `glReadPixels` / `glDrawPixels`, and texture internal formats.  It took us a really, really long time to get that stuff "right" in Mesa, and I'd really hate to complicate that fragile code any further.

Since Mesa and NVIDIA both allow creating a real, conformant ES 1.x context... why not just use the real thing?

Perksey:
My proposed version will likely be heavily based on OES_fixed_point in
writing so it might be worth looking at how that’s handled today.

Perksey:
Haven't spent much more time looking at this, personally don't have much motivation to do so given that OES_single_precision and OES_fixed_point cover most of NV_ES_1_1_compatibility, just with the OES suffix on the functions.

If someone still wanted to create a spec for NV_ES_1_1_compatibility I would use those two extensions as a starting point.

lygstate:
> Doing an extension requires more than just adding the functions to the XML, it also needs to be documented (see https://www.khronos.org/registry/OpenGL/index_gl.php). You would probably make quicker progress making this a vendor or EXT extension rather than an ARB extension. ARB extensions require ratification from the Khronos board, as well as 3+ implementations and CTS.
> 
> Who are you hoping will implement this? NVIDIA have something similar called GL_NV_ES1_1_compatibility, which was never published. That extension seems to expose a lot more functions than the 4 you have exposed here.
> 
> ```c++
> GLAPI void GLAPIENTRY glAlphaFuncx (GLenum func, GLclampx ref);
> GLAPI void GLAPIENTRY glClearColorx (GLclampx red, GLclampx green, GLclampx blue, GLclampx alpha);
> GLAPI void GLAPIENTRY glClearDepthx (GLclampx depth);
> GLAPI void GLAPIENTRY glClipPlanef (GLenum plane, const GLfloat *equation);
> GLAPI void GLAPIENTRY glClipPlanex (GLenum plane, const GLfixed *equation);
> GLAPI void GLAPIENTRY glColor3x (GLfixed red, GLfixed green, GLfixed blue);
> GLAPI void GLAPIENTRY glColor4x (GLfixed red, GLfixed green, GLfixed blue, GLfixed alpha);
> GLAPI void GLAPIENTRY glDepthRangex (GLclampx zNear, GLclampx zFar);
> GLAPI void GLAPIENTRY glFogx (GLenum pname, GLfixed param);
> GLAPI void GLAPIENTRY glFogxv (GLenum pname, const GLfixed *params);
> GLAPI void GLAPIENTRY glFrustumf (GLfloat left, GLfloat right, GLfloat bottom, GLfloat top, GLfloat zNear, GLfloat zFar);
> GLAPI void GLAPIENTRY glFrustumx (GLfixed left, GLfixed right, GLfixed bottom, GLfixed top, GLfixed zNear, GLfixed zFar);
> GLAPI void GLAPIENTRY glGetClipPlanef (GLenum plane, GLfloat *eqn);
> GLAPI void GLAPIENTRY glGetClipPlanex (GLenum plane, GLfixed *eqn);
> GLAPI void GLAPIENTRY glGetFixedv (GLenum pname, GLfixed *params);
> GLAPI void GLAPIENTRY glGetLightxv (GLenum light, GLenum pname, GLfixed *params);
> GLAPI void GLAPIENTRY glGetMaterialxv (GLenum face, GLenum pname, GLfixed *params);
> GLAPI void GLAPIENTRY glGetTexEnvxv (GLenum env, GLenum pname, GLfixed *params);
> GLAPI void GLAPIENTRY glGetTexParameterxv (GLenum target, GLenum pname, GLfixed *params);
> GLAPI void GLAPIENTRY glLightModelx (GLenum pname, GLfixed param);
> GLAPI void GLAPIENTRY glLightModelxv (GLenum pname, const GLfixed *params);
> GLAPI void GLAPIENTRY glLightx (GLenum light, GLenum pname, GLfixed param);
> GLAPI void GLAPIENTRY glLightxv (GLenum light, GLenum pname, const GLfixed *params);
> GLAPI void GLAPIENTRY glLineWidthx (GLfixed width);
> GLAPI void GLAPIENTRY glLoadMatrixx (const GLfixed *m);
> GLAPI void GLAPIENTRY glMaterialx (GLenum face, GLenum pname, GLfixed param);
> GLAPI void GLAPIENTRY glMaterialxv (GLenum face, GLenum pname, const GLfixed *params);
> GLAPI void GLAPIENTRY glMultiTexCoord1x (GLenum target, GLfixed s);
> GLAPI void GLAPIENTRY glMultiTexCoord1xv (GLenum target, const GLfixed *v);
> GLAPI void GLAPIENTRY glMultiTexCoord2x (GLenum target, GLfixed s, GLfixed t);
> GLAPI void GLAPIENTRY glMultiTexCoord2xv (GLenum target, const GLfixed *v);
> GLAPI void GLAPIENTRY glMultiTexCoord3x (GLenum target, GLfixed s, GLfixed t, GLfixed r);
> GLAPI void GLAPIENTRY glMultiTexCoord3xv (GLenum target, const GLfixed *v);
> GLAPI void GLAPIENTRY glMultiTexCoord4x (GLenum target, GLfixed s, GLfixed t, GLfixed r, GLfixed q);
> GLAPI void GLAPIENTRY glMultiTexCoord4xv (GLenum target, const GLfixed *v);
> GLAPI void GLAPIENTRY glMultMatrixx (const GLfixed *m);
> GLAPI void GLAPIENTRY glNormal3x (GLfixed nx, GLfixed ny, GLfixed nz);
> GLAPI void GLAPIENTRY glOrthof (GLfloat left, GLfloat right, GLfloat bottom, GLfloat top, GLfloat zNear, GLfloat zFar);
> GLAPI void GLAPIENTRY glOrthox (GLfixed left, GLfixed right, GLfixed bottom, GLfixed top, GLfixed zNear, GLfixed zFar);
> GLAPI void GLAPIENTRY glPointParameterx (GLenum pname, GLfixed param);
> GLAPI void GLAPIENTRY glPointParameterxv (GLenum pname, const GLfixed *params);
> GLAPI void GLAPIENTRY glPointSizex (GLfixed size);
> GLAPI void GLAPIENTRY glPolygonOffsetx (GLfixed factor, GLfixed units);
> GLAPI void GLAPIENTRY glRotatex (GLfixed angle, GLfixed x, GLfixed y, GLfixed z);
> GLAPI void GLAPIENTRY glSampleCoveragex (GLclampx value, GLboolean invert);
> GLAPI void GLAPIENTRY glScalex (GLfixed x, GLfixed y, GLfixed z);
> GLAPI void GLAPIENTRY glTexEnvx (GLenum target, GLenum pname, GLfixed param);
> GLAPI void GLAPIENTRY glTexEnvxv (GLenum target, GLenum pname, const GLfixed *params);
> GLAPI void GLAPIENTRY glTexParameterx (GLenum target, GLenum pname, GLfixed param);
> GLAPI void GLAPIENTRY glTexParameterxv (GLenum target, GLenum pname, const GLfixed *params);
> GLAPI void GLAPIENTRY glTranslatex (GLfixed x, GLfixed y, GLfixed z);
> ```

I am curious where the list comes from

pdaniell-nv:
Those are the entry points the NVIDIA driver exposes to support OpenGL ES 1.1 and pass its conformance tests.

lorn10:
It should be noted that mentioned **GL_NV_ES1_1_compatibility** extension is now available also in Mesa. :+1: 

I can confirm this for Mesa 22.3.0-devel (git-ef27399 2022-08-27). More information can be found at the corresponding MR [!18217](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/18217)

mirh:
This was [recently](https://gitlab.winehq.org/wine/wine/-/merge_requests/9464) added to wine, for some reason.

PR #456, opened 2021-01-27
GLX_SGI_video_sync: Note the interaction with ARB_framebuffer_object

nwnk:
This is somewhat trivial since it just says the behavior is undefined,
but implementations do diverge in the wild so better to put up some
caution tape.


pdaniell-nv:

PR #459, opened 2021-02-08
Added SizedInternalFormat enum group for functions that need explicitly sized formats

NogginBops:
This PR adds a new enum group `SizedInternalFormat` that is used to more specifically group the enums that are valid in calls to functions such as `glTexStorage2D` etc.

I could find documentation for most of the functions and enums, but `EXT_texture_compression_latc` doesn't specify if it can be passed to functions like `glTexStorage2D` or not, so I decided to not add the `SizedInternalFormat` group to those.
```
GL_COMPRESSED_LUMINANCE_LATC1_EXT
GL_COMPRESSED_SIGNED_LUMINANCE_LATC1_EXT
GL_COMPRESSED_LUMINANCE_ALPHA_LATC2_EXT
GL_COMPRESSED_SIGNED_LUMINANCE_ALPHA_LATC2_EXT
```

I also added the `InternalFormat` group to a few enums and commands where they where missing. (e.g. `glMultiTexBufferEXT`)

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -5817,7 +5817,7 @@ typedef unsigned int GLhandleARB;
         <enum value="0x2A02" name="GL_POLYGON_OFFSET_LINE" group="GetPName,EnableCap"/>
         <enum value="0x2A02" name="GL_POLYGON_OFFSET_LINE_NV"/>
             <unused start="0x2A03" end="0x2A09" comment="Unused for PolygonOffset"/>
-        <enum value="0x2A10" name="GL_R3_G3_B2" group="InternalFormat"/>
+        <enum value="0x2A10" name="GL_R3_G3_B2" group="InternalFormat,SizedInternalFormat"/>
             <unused start="0x2A11" end="0x2A1F" comment="Unused for InternalFormat"/>
         <enum value="0x2A20" name="GL_V2F" group="InterleavedArrayFormat"/>
         <enum value="0x2A21" name="GL_V3F" group="InterleavedArrayFormat"/>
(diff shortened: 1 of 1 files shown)


xml/gl.xml
```diff
@@ -12942,15 +12942,15 @@ typedef unsigned int GLhandleARB;
         <command>
             <proto>void <name>glClearBufferData</name></proto>
             <param group="BufferStorageTarget"><ptype>GLenum</ptype> <name>target</name></param>
-            <param group="InternalFormat"><ptype>GLenum</ptype> <name>internalformat</name></param>
+            <param group="SizedInternalFormat"><ptype>GLenum</ptype> <name>internalformat</name></param>
```
Perksey, line 12945: Parameter group changes are breaking changes. I guess in this instance it could be justified in that all the enumerators that can be passed into this function have been marked with this new group, and that the groups aren't used for Khronos purposes so I suppose this is fine.

I don't know how breaking this will be for other downstream consumers though that generate strongly-typed enums using these attributes - perhaps it could be a good idea to establish a proper policy on these kinds of changes in the future.

Perksey (approved):
LGTM

xml/gl.xml
Lokathor: If passing any InternalFormat before was actually wrong, and then SizedInternalFormat actually does the right thing, then this is technically a break but more importantly this is a bugfix, which generally takes priority over absolutely never doing a break.

Speaking as a person who spent all last week rebuilding their Rust GL bindings generator from scratch: I have strong typing on the GLenum / GLbitmask split, but not on the groups within GLenum (those are all just type aliases). When I asked the other guy who also uses my bindings directly they actually voted for "just be as direct as possible, don't try to fix the typing on anything if you can avoid it".

Perksey: Yeah I've generally taken the same approach as well, but there's no established policy for breaks like this, especially considering it's not breaking for Khronos directly but breaking for a lot of their downstream consumers, including members such as Google and its ANGLE project (cc @null77)

pdaniell-nv:

oddhack:
> @oddhack this is approved to merge.

Merging, although given comments about potentially breaking ANGLE, perhaps it would be wise to identify a person on that project to sign off on these changes. This is something I was concerned about when the groups started getting actively updated - we have no ability to do more than the most basic sanity checks on proposed changes to the groups, since we don't use them, they don't affect us, and we don't really know who's using them downstream.

frederikja163:
> > @oddhack this is approved to merge.
> 
> Merging, although given comments about potentially breaking ANGLE, perhaps it would be wise to identify a person on that project to sign off on these changes. This is something I was concerned about when the groups started getting actively updated - we have no ability to do more than the most basic sanity checks on proposed changes to the groups, since we don't use them, they don't affect us, and we don't really know who's using them downstream.

Sorry to revive this, i was just interested in wether or not you would be interested in a list of all projects that use the groups and/or other of the smaller features that arent used internally. This could include things such as the len attributes and more. Im thinking making an issue and pinning it, here projects can then comment if they want to get added to said list.

The list could be used as a reference for what projects use different features and talk with atleast some of them before making changes plus it would better allow khronos to know who is making changes and what their goal is. We could also add some additional contact information/contact people per project using the spec.

If khronos is interested in this i can go ahead and make said issue and i can also be responsible for updating it, at the very least in the near future.

SunSerega:
> Im thinking making an issue and pinning it, here projects can then comment if they want to get added to said list.

I think discussions are, by design, more suited for this.
And to make people see it info when making issues - there are issue templates. You can add something like this to the start:
```
<!--- People that might be affected by XML change, consider mentioning so they can approve --->
[ANGLE](...): @...
Other: @...
```

Perksey:
We should move this discussion to its own issue.

PR #461, opened 2021-03-04
Len parameters

frederikja163:
Fixes some len parameters, I went through all the lens that have a compsize and they should now all be pretty consistent such that none of them are not actually computing sizes.

Here are a few outliers that either had very difficult to find documentation or which i couldnt find the specific documentation i was looking for.
```
glGetUniformIndices
glGetnMapdv - No documentation??
glMultiDrawArraysIndirectCount
glMultiDrawElementsIndirectCount
glMulticastScissorArrayvNVX
glMulticastViewportArrayvNVX
glScissorExclusiveArrayvNV
glTransformFeedbackAttribsNV
glVariantbvEXT
glVertexArrayRangeNV
glWindowRectanglesEXT
```

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -27440,20 +27440,20 @@ typedef unsigned int GLhandleARB;
             <proto>void <name>glScissorArrayv</name></proto>
             <param><ptype>GLuint</ptype> <name>first</name></param>
             <param><ptype>GLsizei</ptype> <name>count</name></param>
-            <param len="COMPSIZE(count)">const <ptype>GLint</ptype> *<name>v</name></param>
+            <param len="count * 4">const <ptype>GLint</ptype> *<name>v</name></param>
         </command>
         <command>
             <proto>void <name>glScissorArrayvNV</name></proto>
             <param><ptype>GLuint</ptype> <name>first</name></param>
             <param><ptype>GLsizei</ptype> <name>count</name></param>
-            <param len="COMPSIZE(count)">const <ptype>GLint</ptype> *<name>v</name></param>
+            <param len="count * 4">const <ptype>GLint</ptype> *<name>v</name></param>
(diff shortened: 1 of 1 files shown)


Lokathor:
many other places also use compsize(arg), why are we removing it here and not those other locations?

frederikja163:
In some places compsize needs to be there, im just removing it here because this is a place where it dosent compute the size but rather it just uses the size directly. This would also make the method fit with the non-dlss version `glBufferSubData`

Perksey (approved):
LGTM (I also can't stress enough that none of the len stuff is well defined and if you're looking for correctness to this extent then you should probably not be doing whatever you're doing! If you'd like a spec with more info I hear the Mesa guys have a nice one going with much more info)

xml/gl.xml
```diff
@@ -27440,20 +27440,20 @@ typedef unsigned int GLhandleARB;
             <proto>void <name>glScissorArrayv</name></proto>
             <param><ptype>GLuint</ptype> <name>first</name></param>
             <param><ptype>GLsizei</ptype> <name>count</name></param>
-            <param len="COMPSIZE(count)">const <ptype>GLint</ptype> *<name>v</name></param>
+            <param len="count * 4">const <ptype>GLint</ptype> *<name>v</name></param>
```
Perksey, line 27443: Awesomeee!

Perksey (approved):

pdaniell-nv:
@frederikja163 did you mean to close this?

ggrzybow:
> Here are a few outliers that either had very difficult to find documentation or which i couldnt find the specific documentation i was looking for.
> `glMultiDrawElementsIndirectCount`

It's documented in the [spec PDF](https://github.com/KhronosGroup/OpenGL-Registry/blob/main/specs/gl/glspec46.core.pdf), but not in the registry.

PR #462, opened 2021-03-06
Fix non-standard array declaration in XML

### LINKED ISSUE #455: glPathGlyphIndexRangeNV has an *old*? syntax for constant size arrays '[2]'
In the xml file `glPathGlyphIndexRangeNV` has a unique way of defining an array of constant length, Should this rather use the Len attribute? Or would such a change break anything on the C side, I know that it would be really handy for our bindings. Im happy to implement the change myself aswell.

Perksey:
Closes #455


Perksey:
Note: did we want to make the GLuint const as in other functions?

pdaniell-nv:
> Note: did we want to make the GLuint const as in other functions?

Not in this case since it's an "out".

Perksey:
> Not in this case since it's an "out".

Ah didn't realize. In which case should be good to go.

pdaniell-nv:

Issue #463, opened 2021-03-26
[xml] functions missing from EXT_gpu_shader4


eddiejames:
Hi, the following function definitions seem to be missing from the XML for EXT_gpu_shader4:

```void VertexAttribI1iEXT(uint index, int x);
void VertexAttribI2iEXT(uint index, int x, int y);
void VertexAttribI3iEXT(uint index, int x, int y, int z);
void VertexAttribI4iEXT(uint index, int x, int y, int z, int w);

void VertexAttribI1uiEXT(uint index, uint x);
void VertexAttribI2uiEXT(uint index, uint x, uint y);
void VertexAttribI3uiEXT(uint index, uint x, uint y, uint z);
void VertexAttribI4uiEXT(uint index, uint x, uint y, uint z, uint w);

void VertexAttribI1ivEXT(uint index, const int *v);
void VertexAttribI2ivEXT(uint index, const int *v);
void VertexAttribI3ivEXT(uint index, const int *v);
void VertexAttribI4ivEXT(uint index, const int *v);

void VertexAttribI1uivEXT(uint index, const uint *v);
void VertexAttribI2uivEXT(uint index, const uint *v);
void VertexAttribI3uivEXT(uint index, const uint *v);
void VertexAttribI4uivEXT(uint index, const uint *v);

void VertexAttribI4bvEXT(uint index, const byte *v);
void VertexAttribI4svEXT(uint index, const short *v);
void VertexAttribI4ubvEXT(uint index, const ubyte *v);
void VertexAttribI4usvEXT(uint index, const ushort *v);

void VertexAttribIPointerEXT(uint index, int size, enum type, sizei stride, const void *pointer);

void GetVertexAttribIivEXT(uint index, enum pname, int *params);
void GetVertexAttribIuivEXT(uint index, enum pname, uint *params);
```

They are instead listed under NV_vertex_program4. This makes sense according to the extension specification, but the extension also notes: `(note:  all these functions are shared with the EXT_gpu_shader4 extension.)`

So I think they should be listed in both places? Not sure how the XML handles that. This is a problem because some drivers don't list NV_vertex_program4 as supported even though they support EXT_gpu_shader4 and the functions above.

oddhack:
That seems correct. The worst consequence of duplicating them in the EXT is that the prototypes may move inside the generated headers, since the header generator is dependency-based. Would you like to submit a PR to this effect, preferably with a comment= in the new \<require> block explaining why the duplication?

PR #464, opened 2021-03-26
Add missing glVertexAttribI* functions in EXT_gpu_shader4

eddiejames:
These functions were previously listed only in NV_vertex_program4, but
are described in both extension specifications. The NV_vertex_program4
extension explicitly states that they are shared with EXT_gpu_shader4,
so the xml should reflect this.

Fixes https://github.com/KhronosGroup/OpenGL-Registry/issues/463


oddhack:
Looks fine to me.

pdaniell-nv:
> Looks fine to me.

@oddhack thanks for taking a look. This is approved to merge.

PR #465, opened 2021-04-09
Remove groups

### LINKED ISSUE #335: XML Format Consultation: Merging <groups> and <enums>

Perksey:
Closes #335


pdaniell-nv:
@oddhack if you are happy with this, it is approved to merge.

PR #468, opened 2021-05-07
 Add the MESA_bgra GLES extension

gerddie:
This extensions adds the pairs BGRA_EXT/RGBA8 and BGR_EXT/RGB8
to the accepted combinations for format/internalformat in the
TexImage and TexSubImage calls. This is of interest in virtualized
enviornments where one wants to emulate a subset of OpenGL in a guest
that runs on a host that supports only GLES.

An implementation is currently about to be merged in mesa: 
  https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/10613

PR #469, opened 2021-05-07
Fixup EXT_texture_sRGB_RG8.txt in glext.php

gerddie:
It seems that glext.php was not properly re-generated,  so that the entry was missing. 

(Noted while submitting https://github.com/KhronosGroup/OpenGL-Registry/pull/468)


oddhack:
Best to let me do the generated files when an extension is actually added (also true of the generated headers). Otherwise it tends to cause conflicts if we accept several extensions at the same time.

PR #471, opened 2021-05-11
Remove 'u' from NV_shader_atomic_uint64 GLASM option

ReinUsesLisp:
Rename the GLASM option `NV_shader_atomic_uint64` to `NV_shader_atomic_int64` (remove the 'u' in 'uint64').

The implementation of this extension checks for `NV_shader_atomic_int64` and signals `uint64` as invalid. The extension itself is named without the 'u', and applications relying on this would break if the driver is changed, so I assume it's a typo.

I think changing the spec to match the current behavior is the path with less friction. I considered opening an issue about this, but opted to open this pull request given the simplicity of the change. Let me know if I missed something so it can be amended.

If the approach here is wrong, please close this pull request.

### DIFF
--- a/extensions/NV/NV_shader_atomic_int64.txt
+++ b/extensions/NV/NV_shader_atomic_int64.txt
@@ -23,8 +23,8 @@ Status
 
 Version
 
-     Last Modified Date:	March 13, 2014
-     NVIDIA Revision:		2
+     Last Modified Date:	May 11, 2021
+     NVIDIA Revision:		3
 
 Number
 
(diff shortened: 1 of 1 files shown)


oddhack:
@pdaniell-nv this is something for NVIDIA to sign off on.

nvpbrown:
@pdaniell-nv pointed me at this.  The proposed edits look fine to me; it appears that the original language had a typo that probably replicated via cut-and-paste.  The only issue I have is that the edit should also include a new revision number in the document header and a new entry in the revision history section of the document.

Thanks for finding and reporting this issue.

-- Pat

extensions/NV/NV_shader_atomic_int64.txt
```diff
@@ -183,6 +183,11 @@ Issues
 
 Revision History
 
+    Revision 3, May 11, 2021
+
+      Rename assembly option from NV_shader_atomic_uint64 to
```
nvpbrown, line 188: I'd suggest:

> Fix typos in the NV_gpu_program5 assembly option name, which should have been specified as "NV_shader_atomic_int64".

to clarify that you're fixing a bug in the spec rather than making a functional change.  Otherwise, looks great - thanks.

PR #473, opened 2021-05-24
Add len parameters to char* methods

frederikja163:
I noticed some methods with char*'s didnt have len attributes. This pr fixes almost all of them.

https://www.khronos.org/registry/OpenGL/extensions/QCOM/QCOM_extended_get2.txt
glExtGetProgramBinarySourceQCOM is missing a len attribute for parameter 'source'
Didn't find a direct message in the documentation. But based on the parameter names, I would say this is fixed correctly.

https://www.khronos.org/registry/OpenGL/extensions/NV/NV_transform_feedback.txt
glGetActiveVaryingNV has a 'COMPSIZE' string length for parameter 'name'!
`The actual number of characters written into <name>, excluding
    the null terminator, is returned in <length>` Hence this should not be a compsize, but rather just point to the length parameter.

https://www.khronos.org/registry/OpenGL/extensions/ANGLE/ANGLE_translated_shader_source.txt
glGetTranslatedShaderSourceANGLE is missing a len attribute for parameter 'source'
Didn't find a direct message in the documentation. But based on the parameter names, I would say this is fixed correctly.

https://www.khronos.org/registry/OpenGL/extensions/INTEL/INTEL_performance_query.txt
glGetPerfCounterInfoINTEL is missing a len attribute for parameter 'counterDesc'
glGetPerfCounterInfoINTEL is missing a len attribute for parameter 'counterName'
glGetPerfQueryInfoINTEL is missing a len attribute for parameter 'queryName'
These have been fixed base off of code samples in the documentation linked above.

glGetPerfQueryIdByNameINTEL is missing a len attribute for parameter 'queryName'
This one is documented in the link above with the other intel commands. But it dosent seem like it actually has a length. This makes this parameter the only char* left (from what i've found) that dosent have a length parameter.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -11775,7 +11775,7 @@ typedef unsigned int GLhandleARB;
             <proto>void <name>glExtGetProgramBinarySourceQCOM</name></proto>
             <param class="program"><ptype>GLuint</ptype> <name>program</name></param>
             <param group="ShaderType"><ptype>GLenum</ptype> <name>shadertype</name></param>
-            <param><ptype>GLchar</ptype> *<name>source</name></param>
+            <param len="length"><ptype>GLchar</ptype> *<name>source</name></param>
             <param><ptype>GLint</ptype> *<name>length</name></param>
         </command>
         <command>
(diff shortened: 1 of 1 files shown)


Perksey (approved):

pdaniell-nv:

frederikja163:
After speaking with noggin @NogginBops it seems like the spec is generally more consistent if we interpret the len attributes as the input length of the buffer rather than the output length of the string. The applicable methods for this has been change accordingly. @pdaniell-nv

pdaniell-nv:
@frederikja163 your update looks good. @oddhack this is approved to merge.

PR #477, opened 2021-07-05
Distribute group attribute from <enums> tags to individual <enum> tags.

NogginBops:
The idea with this PR is to fix an inconsistency where most `<enum>` tags within an `<enums>` tag with a `group` attribute also contain a `group` attribute with the same group.
Eg.
```xml
<enums namespace="GL" group="AttribMask" type="bitmask">
        <enum value="0x00000001" name="GL_CURRENT_BIT" group="AttribMask"/>
        ...
```
The group `"AttribMask"` is annotated both on the `enums` tag as well as the individual `enum` tags.

But there are 5 enum groups where this isn't the case.
```
group="RegisterCombinerPname"
group="PathRenderingMaskNV"
group="PathRenderingTokenNV"
group="SpecialNumbers"
group="ContainerType"
```

What this PR does currently:
- [x] Distributed group="RegisterCombinerPname" from the `<enums>` tag to the individual `<enum>` tags.
- [ ] Distributed group="PathRenderingMaskNV" from the `<enums>` tag to the individual `<enum>` tags.
- [ ] Distributed group="PathRenderingTokenNV" from the `<enums>` tag to the individual `<enum>` tags.
- [x] Distributed group="SpecialNumbers" from the `<enums>` tag to the individual `<enum>` tags.
- [x] Distributed group="ContainerType" from the `<enums>` tag to the individual `<enum>` tags.

There are two interesting questions relating to the remaining groups:

First, both `"PathRenderingMaskNV"` and `"PathRenderingTokenNV"` are defined on `enums` tags while all of the `enum` tags inside have their own appropriate groups, so it wouldn't make sense to add these enum groups as they don't make sense (they are only referenced in some `<unused>` tags for `"PathRenderingTokenNV"`).

Second, the `"SpecialNumbers"` group doesn't fit the same mold as all of the other groups (unusual mix of types), ~~so I also think that it wouldn't really make sense to distribute this into the `<enum>` tags.~~ we could easily distribute it to the `<enum>` tags, and I think that is the solution, for some reason it initially felt weird to me.

As the `<enums group="">` attributes basically become useless with this PR I wonder if it wouldn't be worth it to actually remove these attributes. 
The only concern I can really see is that it's likely people have special logic for the `"SpecialNumbers"` group that would break if it where removed, but it would be good if people could weigh in on this.
Removing the `group` attribute from `<enums>` tags isn't that important and I would be fine with just leaving them, but any sort of tidiness is a win with `gl.xml` I feel.

Pinging some people how might have input on this (taken from #335 as there is not a ping-list for people that are affected by these types of changes):
@frederikja163 @Perksey @oddhack @3b @null77 
Sorry if you didn't want to get pinged 🙂

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -2939,11 +2939,11 @@ typedef unsigned int GLhandleARB;
         <enum value="0x8575" name="GL_INTERPOLATE" group="RegisterCombinerPname"/>
         <enum value="0x8575" name="GL_INTERPOLATE_ARB" group="RegisterCombinerPname"/>
         <enum value="0x8575" name="GL_INTERPOLATE_EXT" group="RegisterCombinerPname"/>
-        <enum value="0x8576" name="GL_CONSTANT" group="PathGenMode" group="RegisterCombinerPname"/>
+        <enum value="0x8576" name="GL_CONSTANT" group="RegisterCombinerPname,PathGenMode"/>
         <enum value="0x8576" name="GL_CONSTANT_ARB" group="RegisterCombinerPname"/>
         <enum value="0x8576" name="GL_CONSTANT_EXT" group="RegisterCombinerPname"/>
         <enum value="0x8576" name="GL_CONSTANT_NV" group="RegisterCombinerPname"/>
-        <enum value="0x8577" name="GL_PRIMARY_COLOR" group="PathColor" group="RegisterCombinerPname"/>
+        <enum value="0x8577" name="GL_PRIMARY_COLOR" group="RegisterCombinerPname,PathColor"/>
         <enum value="0x8577" name="GL_PRIMARY_COLOR_ARB" group="RegisterCombinerPname"/>
         <enum value="0x8577" name="GL_PRIMARY_COLOR_EXT" group="RegisterCombinerPname"/>
         <enum value="0x8578" name="GL_PREVIOUS" group="RegisterCombinerPname"/>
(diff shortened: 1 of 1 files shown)


Perksey (approved):
Seems fine, in line with the RelaxNG spec, and consistent with the rest of the registry.

Perksey:
> As the <enums group=""> attributes basically become useless with this PR I wonder if it wouldn't be worth it to actually remove these attributes.

Oh yeah this would be breaking. I think we should keep it in for historical reasons. I doubt there'll ever be any more `enums` with a `group` attribute going forward, but as you said people will likely be using this attribute meaningfully as with other attributes in the spec.

The `groups` blocks were removed because they were significantly outdated and to encourage extension authors to declare their groups as they make extensions (because they're not as out-of-the-way as they used to be, there's little reason not to define a group anymore) whereas I don't think the `enums`-level group has such issue nor is it as meaningful as `groups` (and the new `enum`-level `group`)

TL;DR yes we could remove it, but I don't think we should as it's not hurting anybody.

Perksey:
Also you do make a good point about `SpecialNumbers` - I don't see any other way to define this without the `enums`-level group as the XML spec has no concept of `const` (and if we gave it one it'd be even more breaking) because all `enum`s are added to the header as macros, so they'd be functionally equivalent anyway.

Too breaking for too little gain I think!

NogginBops:
Regarding `SpecialNumbers`, I've changed my opinion in favor of actually distributing it into the `<enum>` tags.
This makes it possible to ignore the `<enums group="">` attribute entirely without any loss in functionality (the two remaining groups `"PathRenderingMaskNV"` and `"PathRenderingTokenNV"` don't provide any functionality and should probably be removed).

pdaniell-nv:
@NogginBops is this still a work in progress?

NogginBops:
All of the changes I want to make are done. 
The groups `"PathRenderingMaskNV"` and `"PathRenderingTokenNV"` shouldn't matter here as they are useless in practice.
The question about what to about the `<enums groups="">` attributes is still an open question.

I wouldn't have any problems merging this as it is.
By merging this PR consumers can effectively ignore the `<enums groups="">` attributes without any loss in functionality, which is nice.

Perksey:
I think leaving `<enums group=""` is the best thing to do here

oddhack:
@pdaniell-nv this looks fine to me.

PR #480, opened 2021-08-09
Glxml fixes

joakim-arm:
Some minor fixes in gl.xml:

### Add `GL_RENDERBUFFER` to the TextureTarget group
The parameter `target` of the `glGetInternalformati*` functions belongs to the `TextureTarget` group. Since `GL_RENDERBUFFER` is a valid value for `target` in those functions, it needs to be added to the `TextureTarget` group.

### Define groups for some enums
 * `GL_FETCH_PER_SAMPLE_ARM` (`EnableCap`, `GetPName`)
 * `GL_FRAGMENT_SHADER_FRAMEBUFFER_FETCH_MRT_ARM` (`GetPName`)

### Change group of parameter `format` in `glCompressed*TexSubImage*` functions
The group of `format` in the `glCompressed*TexSubImage*` functions was set to `PixelFormat`. However, the correct group seems to be `InternalFormat`, which is containing the valid values for that parameter.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -5007,10 +5007,10 @@ typedef unsigned int GLhandleARB;
     <enums namespace="GL" start="0x8D60" end="0x8D6F" vendor="OES">
         <enum value="0x8D60" name="GL_TEXTURE_GEN_STR_OES"/>
         <enum value="0x8D61" name="GL_HALF_FLOAT_OES"/>
-        <enum value="0x8D62" name="GL_RGB565_OES"/>
-        <enum value="0x8D62" name="GL_RGB565"/>
+        <enum value="0x8D62" name="GL_RGB565_OES" group="InternalFormat,SizedInternalFormat"/>
+        <enum value="0x8D62" name="GL_RGB565" group="InternalFormat,SizedInternalFormat"/>
             <unused start="0x8D63" vendor="OES" comment="Was GL_TEXTURE_IMMUTABLE_LEVELS in draft ES 3.0 spec"/>
-        <enum value="0x8D64" name="GL_ETC1_RGB8_OES"/>
+        <enum value="0x8D64" name="GL_ETC1_RGB8_OES" group="InternalFormat,SizedInternalFormat"/>
         <enum value="0x8D65" name="GL_TEXTURE_EXTERNAL_OES"/>
         <enum value="0x8D66" name="GL_SAMPLER_EXTERNAL_OES"/>
         <enum value="0x8D67" name="GL_TEXTURE_BINDING_EXTERNAL_OES"/>
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -9180,7 +9180,7 @@ typedef unsigned int GLhandleARB;
             <param group="CheckedInt32"><ptype>GLint</ptype> <name>level</name></param>
             <param group="CheckedInt32"><ptype>GLint</ptype> <name>xoffset</name></param>
             <param><ptype>GLsizei</ptype> <name>width</name></param>
-            <param group="PixelFormat"><ptype>GLenum</ptype> <name>format</name></param>
+            <param group="InternalFormat"><ptype>GLenum</ptype> <name>format</name></param>
             <param><ptype>GLsizei</ptype> <name>imageSize</name></param>
             <param len="imageSize">const void *<name>bits</name></param>
         </command>
@@ -9193,7 +9193,7 @@ typedef unsigned int GLhandleARB;
             <param group="CheckedInt32"><ptype>GLint</ptype> <name>yoffset</name></param>
             <param><ptype>GLsizei</ptype> <name>width</name></param>
             <param><ptype>GLsizei</ptype> <name>height</name></param>
-            <param group="PixelFormat"><ptype>GLenum</ptype> <name>format</name></param>
+            <param group="InternalFormat"><ptype>GLenum</ptype> <name>format</name></param>
             <param><ptype>GLsizei</ptype> <name>imageSize</name></param>
             <param len="imageSize">const void *<name>bits</name></param>
         </command>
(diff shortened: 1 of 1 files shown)


xml/gl.xml
```diff
@@ -9180,7 +9180,7 @@ typedef unsigned int GLhandleARB;
             <param group="CheckedInt32"><ptype>GLint</ptype> <name>level</name></param>
             <param group="CheckedInt32"><ptype>GLint</ptype> <name>xoffset</name></param>
             <param><ptype>GLsizei</ptype> <name>width</name></param>
-            <param group="PixelFormat"><ptype>GLenum</ptype> <name>format</name></param>
+            <param group="InternalFormat"><ptype>GLenum</ptype> <name>format</name></param>
```
Perksey, line 9183: It's awesome to see that vendors are starting to pitch in on the groups!

This makes sense, but I think we need to establish a breaking change policy for groups going forwards. First Google and now Arm, it looks like even though Khronos themselves don't use the groups they're becoming of increasing importance among the members therein.

Other than that fine with this.

Perksey (approved):

xml/gl.xml
Perksey: If you're able to provide details on your use case for the groups this'd help, but no worries if not. I'll open a separate issue for establishing this breaking change policy.

joakim-arm: It's for pretty-printing traced API calls so that GLenum/GLbitfields appear with their _name_ instead of their _integer_ value. gl.xml is used for generating a lookup table that makes this conversion possible.

GLenum value to name is not a one-to-one mapping (e.g. `0` can be `GL_POINTS` or `GL_ZERO`). The parameter groups make it possible to retrieve the relevant set of GLenum names for a specific parameter of a specific function.

xml/gl.xml
```diff
@@ -5007,8 +5007,8 @@ typedef unsigned int GLhandleARB;
     <enums namespace="GL" start="0x8D60" end="0x8D6F" vendor="OES">
         <enum value="0x8D60" name="GL_TEXTURE_GEN_STR_OES"/>
         <enum value="0x8D61" name="GL_HALF_FLOAT_OES"/>
-        <enum value="0x8D62" name="GL_RGB565_OES"/>
-        <enum value="0x8D62" name="GL_RGB565"/>
+        <enum value="0x8D62" name="GL_RGB565_OES" group="InternalFormat,SizedInternalFormat"/>
+        <enum value="0x8D62" name="GL_RGB565" group="InternalFormat,SizedInternalFormat"/>
             <unused start="0x8D63" vendor="OES" comment="Was GL_TEXTURE_IMMUTABLE_LEVELS in draft ES 3.0 spec"/>
         <enum value="0x8D64" name="GL_ETC1_RGB8_OES"/>
```
NogginBops, line 5013: This should probably also be marked with `SizedInternalFormat` while we are at it.

https://www.khronos.org/registry/OpenGL/extensions/OES/OES_compressed_ETC1_RGB8_texture.txt

xml/gl.xml
```diff
@@ -4948,7 +4948,7 @@ typedef unsigned int GLhandleARB;
         <enum value="0x8D40" name="GL_FRAMEBUFFER" group="ObjectIdentifier,FramebufferTarget,CheckFramebufferStatusTarget"/>
         <enum value="0x8D40" name="GL_FRAMEBUFFER_EXT"/>
         <enum value="0x8D40" name="GL_FRAMEBUFFER_OES" group="FramebufferTarget"/>
-        <enum value="0x8D41" name="GL_RENDERBUFFER" group="ObjectIdentifier,RenderbufferTarget,CopyImageSubDataTarget"/>
+        <enum value="0x8D41" name="GL_RENDERBUFFER" group="ObjectIdentifier,RenderbufferTarget,CopyImageSubDataTarget,TextureTarget"/>
```
NogginBops, line 4951: I'm not sure on this change, as currently `TextureTarget` kind of unofficially means "enums that you can pass to `glTexImage*`" and `GL_RENDERBUFFER` isn't really in that group.

Generally I think we might need to create more groups than just `TextureTarget` as it's also used for `glBindTexture` which doesn't allow you to pass a proxy texture as a target.
Other examples of functions that use `TextureTarget` but specifies a different valid enum set include `glGetInternalformat` (as pointed out in this PR) and `glGetTexLevelParameter`.

NogginBops: I'm fine with leaving it like this for this PR (I think), but I think we need to start a discussion on whether we should create more groups for this.

xml/gl.xml
joakim-arm: Added

xml/gl.xml
joakim-arm: Agreed. Creating more groups seems to be the best long-term solution. However, glGetInternalformat, glGetTexLevelParameter, and glBindTexture seem all to use different sets of valid values for the target parameters. This means potentially three new groups. In addition, the valid target values of the different functions depend on the API version.

Instead of doing this complex refactoring, I went for the simpler route -- keep the group that glGetInternalformat is already using (TextureTarget) and add GL_RENDERBUFFER to it.

NogginBops: I think the "depending on opengl version" enums is something that should be figured out from the `<feature>` tags.

Having the enums groups be the set of all possibly valid enums that can then be filtered out by version and other factors like extension is going to be the most useful form of these enum groups.

pdaniell-nv: @NogginBops are you okay with this PR? I'm not sure if you're requesting a change, or just pointing out gl.xml is not perfect yet.

NogginBops: As I think the groups discussion will result in us looking at these groups more thoroughly again (and hopefully fix them), I think there is not much harm in merging this. So yes, I think this can be merged as is.

pdaniell-nv:

PR #484, opened 2021-09-22
Finish adding GL_EXT_texture_storage to big-GL

nwnk:
The spec text says this is defined even for GL 1.2, but we don't have a
GL extension number for it and the headers don't mention it. Fix that so
we can extend immutable-format texture image support to GL < 4.2.

### DIFF
--- a/api/GL/glcorearb.h
+++ b/api/GL/glcorearb.h
@@ -4678,6 +4678,41 @@ GLAPI void APIENTRY glFramebufferFetchBarrierEXT (void);
 #define GL_EXT_texture_shadow_lod 1
 #endif /* GL_EXT_texture_shadow_lod */
 
+#ifndef GL_EXT_texture_storage
+#define GL_EXT_texture_storage 1
+#define GL_TEXTURE_IMMUTABLE_FORMAT_EXT   0x912F
+#define GL_ALPHA8_EXT                     0x803C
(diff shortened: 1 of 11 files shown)


nwnk:
I guess ARB_texture_storage does exist, so possibly a better move is to remove the big-gl reference from EXT?

dgkoch:
> I guess ARB_texture_storage does exist, so possibly a better move is to remove the big-gl reference from EXT?

And the ARB version has ES interactions.. They look like 90% or more similar..
I forget the history, but I guess we didn't manage to get the GL and ES groups to quite agree to promote it at the same time, so we couldn't end up making it KHR and having one spec. 

Maybe we should just add a note to both and say for ES use the EXT, and for GL use the ARB?

oddhack:
> Maybe we should just add a note to both and say for ES use the EXT, and for GL use the ARB?

Might be worth checking gpuinfo to see if drivers don't comply with that approach.

There are a couple of other extensions with similar disclaimers, IIRC. Before Khronos, there was no way to have an extension approved by both the Khronos ES WG and the ARB.

dgkoch:
It actually looks like NV/AMD/Intel expose GL_EXT_texture_storage in desktop drivers, so I guess the proposed change here is doing the right thing.

pdaniell-nv:
This is approved to merge, but @oddhack will probably want this PR to be updated to not include the generated files.

PR #495, opened 2022-01-16
Change define WIN32_LEAN_AND_MEAN for windows.h in GCC

Jamaika1:


pdaniell-nv:
Why is this change necessary? I would assume "#define WIN32_LEAN_AND_MEAN 1" works just as well as "#define WIN32_LEAN_AND_MEAN".

Also these edits are to generated files, so you'll need to update the generator instead.

PR #506, opened 2022-05-11
Remove BOM from gl.xml

oddhack:
This was inserted in #343, probably as an accidental side-effect of
someone's editor - it is not mentioned in the PR comments - and is
breaking downstream scripts in vk-gl-cts which consume the file. The file encoding is defined to be UTF-8 in any event.


Perksey (approved):
Apologies for this, looks like the C# application we used to automate the migration to group attributes introduces a BOM by default. Not having this won't break us so should be fine.

PR #504, opened 2022-04-21
Add GL_EXT_shader_samples_identical

jekstrand:
We've been shipping this for 6.5 years in Mesa; it's high time we
actually publish it.

PR #509, opened 2022-05-30
Fix ES extension numbering conflict between GL_EXT_shader_samples_identical and GL_EXT_separate_depth_stencil

janharaldfredriksen-arm:
It looks like https://github.com/KhronosGroup/OpenGL-Registry/pull/504 didn't update registry.py - and then there was a race with GL_EXT_separate_depth_stencil resulting in both taking ES extension number 338.

GL_EXT_separate_depth_stencil  is now not visible from https://www.khronos.org/registry/OpenGL/index_es.php.

This resolves the numbering conflict in favor of GL_EXT_separate_depth_stencil just because that was merged first.

CC: @jekstrand, @oddhack

### DIFF
--- a/extensions/EXT/EXT_shader_samples_identical.txt
+++ b/extensions/EXT/EXT_shader_samples_identical.txt
@@ -30,7 +30,7 @@ Version
 Number
 
     OpenGL Extension #557
-    OpenGL ES Extension #338
+    OpenGL ES Extension #339
 
 Dependencies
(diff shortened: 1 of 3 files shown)


oddhack:
LGTM.

oddhack:
I'm just going to go ahead and merge this since it's purely an infrastructure fix to the registry, not API-related.

PR #514, opened 2022-06-13
Add glsc2 headers to makefile for generation

DeepakRavikumarIMG:
glsc2 headers were missed out from the GLHEADERS in the makefile and as
a result they were not generated.

Fix: Added glsc2 headers to the makefile.


oddhack:
LGTM. @pdaniell-nv I don't recall why we weren't generating these headers in the Makefile but the PR has no effect on them other than changing the date in a comment, so we should accept.

pdaniell-nv (approved):

pdaniell-nv:
I also don't know why these were missing. I guess there have been no consumers of it until now, so it went unnoticed.

PR #518, opened 2022-06-24
[xml] Add missing group "FogCoordSrc"

SunSerega:
Enums of groups `FogMode` and `FogCoordSrc` are accepted by the second parameter of `glFogi`, depending on the first parameter:
https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glFog.xml
Though there is currently no way to specify this relation in the `glFogi` definition...


Perksey (approved):

pdaniell-nv:
Approved to merge. Since this is group property it won't affect working group members.

Issue #481, opened 2021-08-10
Establishing a breaking change policy for groups


Perksey:
The strongly-typed enum groups which were "standardised" in #335 and have been heavily iterated upon for correctness since are garnering more use as time goes on. Google (for ANGLE) and Arm (for a debugging/API trace tool: [more info](https://github.com/KhronosGroup/OpenGL-Registry/pull/480#discussion_r685805570)).

Given Khronos do not use the group information directly themselves, they have often gone forgotten about and whatever a incoming community contribution says, goes. I think given the increasing importance of these groups we should be wary of breaking changes.

In bindings generated from the gl.xml, the groups may be used to define more friendly API signatures. A good example from the Silk.NET library can be found [here](https://github.com/dotnet/Silk.NET/blob/main/src/OpenGL/Extensions/Silk.NET.OpenGL.Extensions.ARB/ArbTextureBufferRange.gen.cs). By changing group information in gl.xml, this causes a break in that library. We have come to accept this, but it's important Khronos knows that implication.

One core tenet of the groups is correctness. They exist to provide information on what enums are valid for a particular parameter. As such, I do think we need some way of changing a parameter to a more correct one but how this looks from a breaking change policy perspective, I don't know. The thing is no group change, even if we did use `enum { ... }`s in C, would be breaking due to how C works.

But yeah, this issue mostly serves as a discussion as I don't know what a good (or any) solution is to this.

Thoughts?
- @joakim-arm @NogginBops @3b @SunSerega @devel0 @tsuoranta @Lucretia @PJB3005 @tksuoran @aloucks @pablode @tkaaad97 @StailGot @mode777 @luca-piccioni (as people who have authored PRs that are exclusively regarding groups in the past)
- @null77 (ANGLE stakeholder)
- @Dav1dde (GLAD)
- @HurricanKai @ThomasMiz (ccs from the Silk.NET project)

Lucretia:
The groups are so important for strongly typed languages, like Ada, where we can have enums which enclose the allowed values.

NogginBops:
You forgot to give @frederikja163 a ping.

Personally I do think that because the enum groupings are very loose as they are today saying that changes to them cannot be made feels like a very bad decision. It will lock the groups into the usability they have today, which albeit useful are not as useful as they could conceptually be.

[...]

PR #520, opened 2022-06-24
[xml] Fixes for defined but unused enum groups

### LINKED ISSUE #360: [xml] Unused groups
Here is list of all groups, which are defined using `group` attribute on `<enum>` tag, but aren't used:
```
ContextFlagMask
ContextProfileMask
SubgroupSupportedFeatures
MapTextureFormatINTEL
TextureSwizzle
TextureCompareMode
PixelTexGenMode
TextureEnvMode
LightEnvModeSGIX
ColorMaterialFace
FeedBackToken
FogMode
GetPixelMap
TextureGenMode
TextureMagFilter
TextureMinFilter
TextureWrapMode
GetConvolutionParameter
ConvolutionBorderModeEXT
PointParameterNameSGIS
TextureFilterFuncSGIS
LightModelColorControl
PixelStoreResampleMode
PixelStoreSubsampleRate
CheckFramebufferStatusTarget
DrawBufferModeATI
FramebufferFetchNoncoherent
```
I didn't check anything other then `group` attributes of `<proto>` and `<param>` tags in `<command>` tag. But if i understand correctly - there is no other use for it. Please correct if i'm wrong.

SunSerega:
fix #360
I mainly stalled in fixing it because of groups like `TextureMinFilter`: It can be passed to, for instance, `glTextureParameteri`, but there is currently no way to denote that in XML - because the same parameter can also take `TextureMagFilter`, `TextureWrapMode`, etc., depending on the `<pname>`.
But since that issue, I updated my data scrappers and generators, so it became easier to separate these cases from actual misdefined groups.

I thought of making a bunch of small PRs, but I really don't like how unreadable that looks on git commit graphs.
So instead, I separated fixes by one or a few related groups. But I can still break this into many PRs if that would be easier to review.


NogginBops:
Not entirely sure why `MaterialFace` got preferred over something like `FaceCullMode`. `glCullFace` taking a `MaterialFace` instead of `FaceCullMode` doesn't make much sense even though the group contains the same enums

SunSerega:
> why `MaterialFace` got preferred

Because it was used in almost all cases, where these enums are accepted. This way it's a minimal change.
Well, If that's not a good reason, how about renaming it to something like `TriangleFace`?

NogginBops:
Sounds like a good idea to me.

Perksey (changes requested):
Some of these groups are unused because we haven't decided on #481 and as a result haven't touched the functions that _could_ use them.

I also don't think we should remove groups purely because they're not declared for usage on a function parameter, they may be used for other reasons: analysers (I know ARM have a tool which consumes the groups used to validate OpenGLES calls and state), storage/fields, etc. 

By removing groups we are actively removing information from this specification, and I'm not sure them looking unused on the surface is a good excuse for that.

SunSerega:
I only removed 5 groups. Look at ` => X` in commits and at descriptions of them:

> #### GetPixelMap => X
> Same as PixelMap, but never used

That is because I've mainly looked at all the enums of these groups and found where they are used, adding `group` attribute to `<param>` or merging them with other groups.
So, at least theoretically, no correct information should be lost in these changes. But I would be glad if anyone could give a fresh look on what exactly is changed.

xml/gl.xml
```diff
@@ -268,13 +268,13 @@ typedef unsigned int GLhandleARB;
     </enums>
 
     <enums namespace="GL" group="TraceMaskMESA" type="bitmask">
-        <enum value="0x0001" name="GL_TRACE_OPERATIONS_BIT_MESA" group="TraceMaskMESA"/>
```
Perksey, line 271: Why has this group been removed? It has not been replaced with an alternative group.

xml/gl.xml
```diff
@@ -564,11 +564,11 @@ typedef unsigned int GLhandleARB;
         <enum value="0x0401" name="GL_FRONT_RIGHT" group="ColorBuffer,DrawBufferMode,ReadBufferMode"/>
         <enum value="0x0402" name="GL_BACK_LEFT" group="ColorBuffer,DrawBufferMode,ReadBufferMode"/>
         <enum value="0x0403" name="GL_BACK_RIGHT" group="ColorBuffer,DrawBufferMode,ReadBufferMode"/>
-        <enum value="0x0404" name="GL_FRONT" group="ColorBuffer,ColorMaterialFace,CullFaceMode,DrawBufferMode,ReadBufferMode,StencilFaceDirection,MaterialFace"/>
```
Perksey, line 567: Seems like a reasonable change but could break a lot downstream. Might want to justify this.

xml/gl.xml
```diff
@@ -5674,11 +5674,11 @@ typedef unsigned int GLhandleARB;
         <enum value="0x8FF9" name="GL_FLOAT16_VEC2_NV"/>
         <enum value="0x8FFA" name="GL_FLOAT16_VEC3_NV"/>
         <enum value="0x8FFB" name="GL_FLOAT16_VEC4_NV"/>
-        <enum value="0x8FFC" name="GL_DOUBLE_VEC2" group="GlslTypeToken,AttributeType,UniformType"/>
```
Perksey, line 5677: Removal of `GlslTypeToken`, replaced with `AttributeType`. Could be breaking downstream, but suitable replacement. Should be fine.

Perksey (approved):
Ok I've cast my eye over it in detail this time and you're right we're not losing too much information, it's just been shuffled in a way that could be quite breaking downstream.

Would be good to get someone from Google and ARM to look over this given they're the known Khronos members to be using groups.

Perksey:
@null77 @joakim-arm Can you look over these changes before approval? It'd be good to use known group users by Khronos members/projects as a litmus test for these sorts of breaking group changes in lieu of #481

xml/gl.xml
SunSerega: Ah, I didn't find this back then:
https://opensource.apple.com/source/X11server/X11server-85/mesa/Mesa-7.0.4/docs/MESA_trace.spec.auto.html
Well, I've added this group back (and also 2 other enums to `GetPName` and `StringName`).
But there is nowhere to use them, because only enums of this extension exist, not functions, so I didn't push this.

I don't think it makes sense to have enums in any binding, when that binding doesn't also expose functions to use them these enums.
If bindings user needs to manually add functions - they should handle related enums too.
So the way to go should be either to add functions and a .txt file to the registry (if they are usable, even if outdated) or remove the enums.

Or, another alternative, if the point of adding enums and only them was to reserve points in enums space, it should instead be done with `<unused/>` tag.

xml/gl.xml
SunSerega: Do you mean the `TriangleFace`?

I initially merged [ColorMaterialFace,CullFaceMode,StencilFaceDirection => MaterialFace](https://github.com/KhronosGroup/OpenGL-Registry/pull/520/commits/55d45650bf9f4914288cc741018f6a63a7cd6d33), because `MaterialFace` was used most often.
But I agree with @NogginBops that it would be better to create a new, more relevant name, as a replacement for these groups duplicating each other.

IMO:
1. Changes in XML don't just immediately appear in bindings. I would expect there to be another layer of checks by the binding maintainer when pulling a new version of this registry. At least because besides actual intended change, there might be a data scrapper quirk, applying these changes in a weird way.
That layer of checks would allow to add something like a group type synonyms, with some kind of deprecated attribute (whatever exists in a language) and then decide when to remove them, or do something else entirely.
Importantly, this is a per binding decision, for which the target language is very relevant.
And if some info isn't needed for any theoretical new binding, if it's only useful for the "compatibility" version of some of the existing bindings - it shouldn't be stored in common XML for all bindings.

2. In the first place, the point of making high-level wrapping for groups of enums is to have diagnostics and IntelliSense suggestions in the target language. So when the group name changes, for the binding user the fix should be trivial.

Of course, it's better to minimize such changes. But I think it is justified here, because removing mess from groups should be prioritized, at least for now.

xml/gl.xml
SunSerega: Also note the [commit](https://github.com/KhronosGroup/OpenGL-Registry/pull/520/commits/1404006e38ea285937592e8157fe21cfabec496b) description, specifically `GL_UNSIGNED_INT_ATOMIC_COUNTER`.
That's the only enum I'm unsure about in this commit.

xml/gl.xml
Perksey: The XML is not just used by bindings, it is used by analysis tools, validation tools, etc. Anything that could possibly break anyone needs to be justified.

SunSerega: I didn't know any of those. Can you please give examples?

Perksey: ARM have an internal validation tool for debugging OpenGLES tools that depend on the groups, and Google’s ANGLE use them in some capacity - likely also validation.

SunSerega: Ah, that's what pings are for, I see...

pdaniell-nv:
Are we still waiting for approval from @null77 and @joakim-arm?

Perksey:
Ideally yes. We can continue without their approval, but ultimately this will likely impact them as group users and given they represent Khronos members/projects, I feel like we should solicit their opinion first!

If we haven't heard back in a little bit (say 2 weeks) then I think we should merge.

joakim-arm:
Looks good, but just one comment:

Removing the group TraceMaskMESA from the bitmask values makes things a bit inconsistent. All other bitmask values are assigned to a group. 

@SunSerega, did you add this group back?

SunSerega:
No, it has an unresolved conversation [here](https://github.com/KhronosGroup/OpenGL-Registry/pull/520#discussion_r906758107).
I encourage other reviewers to also look at unresolved conversations. They are what other people found non-trivial and what I left unresolved because I would appreciate more feedback.

---

I personally think these enums should be removed from XML because there is no documentation/functions in this repo, corresponding to these enums.
Seeing how smoothly #523 went, I guess this kind of change isn't seen as breaking, so I'll go ahead and comment these out.

Perksey:
> I personally think these enums should be removed from XML because there is no documentation/functions in this repo, corresponding to them.
Seeing how smoothly https://github.com/KhronosGroup/OpenGL-Registry/pull/523 went, I guess this kind of change isn't seen as breaking, so I'll go ahead and comment these out.

I don't agree with this, mainly because:

> Possibly fun historical note: early in GL history, when it was still all workstation manufacturers, one of the CAD vendors came up with a "secret" extension. They convinced multiple GL implementers to support it in their drivers, but not to expose it in the GL_EXTENSIONS string. We speculated the purpose was to give them some sort of competitive edge over other CAD vendors, while keeping the driver suppliers from realizing what was going on.
> - oddhack

We shouldn't actively remove information from the registry just because it doesn't look to be used. We have 30 years of history to cover here, can you really say that this information wasn't relevant for the entirety of those 30 years? Moreover, it's unknowable what dependency someone will have taken on any part of the registry, so I don't think it's really great to make these sorts of breaks lightly. If a Khronos member approves the removal of this, given groups aren't WG-controlled we can remove it but only if there is reasonable cause & consensus.

joakim-arm:
> Looks good, but just one comment:

I just wanted to clarify that this change doesn't break anything for our internal tool. However, I am afraid I don't have enough background and resources to comment on other possible implications that this change could introduce.

What Perksey says about covering history makes sense to me.

NogginBops:
I'm fine with these changes too. 
They are going to cause a lot of breaking changes in bindings, but as we are preparing for a new major version, I personally have no objection.

oddhack:
I agree with @Perksey - completely removing enums from the registry affects the Historical Record. If an enum isn't being used by an extension but was still reserved that's useful to document, and nobody needs to be actually processing it if it isn't tagged as part of an extension or core API.

SunSerega:
Ok, then how about defining a special group for such enums, to show they only exist as a legacy?

oddhack:
> Ok, then how about defining a special group/attribute for such enums, to show they only exist as a legacy?
> 
> Note enums in question aren't even reserved in this case, because they are bitmasks.

We don't have any historical preservation need to add *new* information about these enums. I'm not sure why it even comes up for anyone. If there's no dependency path from an extension requiring these values, they should be ignored by whatever is consuming the XML. If the people concerned with the groups collectively agree there's some value to adding it, though, feel free.

NogginBops:
I would also be careful with removing enums, and I don't think there is a need for a special group to mark them. 
Though I would be open to hear why such a group might be useful if you (@SunSerega) want to argue for it.

SunSerega:
I think this has grown a bit out of the scope of this pull, so I'll remove changes to `TraceMaskMESA` and make a separate issue after sorting my thoughts a bit more.

xml/gl.xml
```diff
@@ -267,15 +267,15 @@ typedef unsigned int GLhandleARB;
...
-    </enums>
+	<enums namespace="GL" group="TraceMaskMESA" type="bitmask">
+		<enum value="0x0001" name="GL_TRACE_OPERATIONS_BIT_MESA" group="TraceMaskMESA"/>
+		<enum value="0x0002" name="GL_TRACE_PRIMITIVES_BIT_MESA" group="TraceMaskMESA"/>
+		<enum value="0x0004" name="GL_TRACE_ARRAYS_BIT_MESA" group="TraceMaskMESA"/>
+		<enum value="0x0008" name="GL_TRACE_TEXTURES_BIT_MESA" group="TraceMaskMESA"/>
+		<enum value="0x0010" name="GL_TRACE_PIXELS_BIT_MESA" group="TraceMaskMESA"/>
+		<enum value="0x0020" name="GL_TRACE_ERRORS_BIT_MESA" group="TraceMaskMESA"/>
+		<enum value="0xFFFF" name="GL_TRACE_ALL_BITS_MESA" group="TraceMaskMESA"/>
+	</enums>
```
Perksey, line 278: Spaces > tabs
```suggestion
    <enums namespace="GL" group="TraceMaskMESA" type="bitmask">
        <enum value="0x0001" name="GL_TRACE_OPERATIONS_BIT_MESA" group="TraceMaskMESA"/>
        <enum value="0x0002" name="GL_TRACE_PRIMITIVES_BIT_MESA" group="TraceMaskMESA"/>
        <enum value="0x0004" name="GL_TRACE_ARRAYS_BIT_MESA" group="TraceMaskMESA"/>
        <enum value="0x0008" name="GL_TRACE_TEXTURES_BIT_MESA" group="TraceMaskMESA"/>
        <enum value="0x0010" name="GL_TRACE_PIXELS_BIT_MESA" group="TraceMaskMESA"/>
        <enum value="0x0020" name="GL_TRACE_ERRORS_BIT_MESA" group="TraceMaskMESA"/>
        <enum value="0xFFFF" name="GL_TRACE_ALL_BITS_MESA" group="TraceMaskMESA"/
    </enums>
```

SunSerega: Thx, fixed.

Perksey (approved):

pdaniell-nv:
As far as I can tell, this PR is complete and awaiting sign-off from the OpenGL/ES working group. We'll approve this tomorrow. Please let me know if anyone has objections to this getting merged, otherwise it will likely be approved for merge tomorrow.

Perksey:
Approved for groups.

pdaniell-nv:

PR #524, opened 2022-07-07
Fix incorrect spelling of some tokens in the spec text of GL_EXT_fragment_shading_rate

janharaldfredriksen-arm:
Two mistakes - an "_BIT" had snuck in, and some cases of 1X1 used a lower-case "x".

The new token section was correct and matches the header, so this is purely editorial.


oddhack (approved):
LGTM.

pdaniell-nv:

PR #526, opened 2022-07-15
Add implicit flush and non-listable commands to external objects

cubanismo:
-SignalSemaphoreEXT implicitly flushes command stream

-None of the new external object commands can be included in display lists

### DIFF
--- a/extensions/EXT/EXT_external_objects.txt
+++ b/extensions/EXT/EXT_external_objects.txt
@@ -532,7 +532,8 @@ Additions to Chapter 4 of the OpenGL 4.5 Specification (Event Model)
                                     const GLenum *dstLayouts);
 
         will insert a semaphore signaling operation in the GL command
-        stream.
+        stream, and flush the command stream as if /Flush/ were inserted
+        immediately after the semaphore operation.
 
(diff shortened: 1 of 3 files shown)


zmike:
Technically I think the language has to be stronger for the flush semantics. It's not sufficient to say `as if /Flush/ were inserted immediately after the semaphore operation` because Flush does not provide any guarantee of immediate GPU workload submission, which is what this command requires.

cubanismo:
I took this language straight from ClientWaitSync, which is what the footnote the people who broke Flush forced into the spec recommends applications use to work around Flush now being broken in the GL spec. I'm not sure what stronger language could be used.

zmike:
Something like `and submit the command stream to the GPU before the function returns` should be required here since the extension can't work otherwise.

cubanismo:
I think if that's the only way to describe a flush in the OpenGL specification, the specification is generally broken. Flush is supposed to flush, and I'm not interested in solving that much breakage in this unrelated extension.

zmike:
Yeah fair enough. Anyone interested enough can find the comments here anyway.

pdaniell-nv:
@oddhack this is approved to merge if you're also happy with it.

PR #528, opened 2022-07-21
Change type of 'severities' argument to enum in GetDebugMessageLogAMD

### LINKED ISSUE #516: GetDebugMessageLogAMD takes GLint* argument instead of GLenum*
The function `GetDebugMessageLogAMD` from `GL_AMD_debug_output` is defined as follows: 
```c
uint GetDebugMessageLogAMD(uint count,
                               GLsizei bufsize,
                               GLenum* categories,
                               GLuint* severities,
                               GLuint* ids,
                               GLsizei* lengths, 
                               GLchar* message);
```

Shouldn't the `severities` argument be of type `GLenum`? 
This would match every other definition of this function from other extensions.

As they are `typedef`ed as the same type is it possible to change this?

NogginBops:
Fixes #516 

Not sure if I should add this change to the change log of the extension text.
I also noticed now that the text later refers to `GetDebugMessageLogAMD` with the new signature, so the extension was inconsistent to begin with.

I'm also adding an extra commit where I change instances of `enum` to be `GLenum`, I can easily revert this if desired.


oddhack:
LGTM.

extensions/AMD/AMD_debug_output.txt
```diff
@@ -93,13 +93,13 @@ IP Status
 New Procedures and Functions
     
     void DebugMessageEnableAMD(enum category,
-                               enum severity,
+                               GLenum severity,
```
Perksey, line 96: Extension specifications omit prefixes for readability.

NogginBops:
@oddhack reverted the `enum` to `GLenum` commit.

NogginBops:
I can't figure out from your message if I should revert the changelog change too, but I'm guessing that there is something more as you said I should revert more than one thing. But I'm unsure of what I'm supposed to remove.

oddhack:
> I can't figure out from your message if I should revert the changelog change too, but I'm guessing that there is something more as you said I should revert more than one thing. But I'm unsure of what I'm supposed to remove.

LGTM. I should not have used "revert" in the first sentence, all I meant was I had previously signed off and then replaced that comment after realizing the GL prefix needed to be dropped.

oddhack:
@pdaniell-nv I think this is OK per discussion on #516? Also, if @rpatlega is a WG member from AMD, please let me know so I can add them to the Github team and @-ping them.

rpatlega:
From AMD side, we are good this for merge.

oddhack:
> From AMD side, we are good this for merge.

Thanks. I'll take that as Signed Off given @pdaniell-nv's comment. BTW I sent you an invite to the github team associated with the OpenGL repos.

PR #530, opened 2022-07-27
Add missing len attributes to GL commands

laranzu:
Add missing "len=" attribs to about ten XML core API command definitions
Correct one "len=" attrib from unnecessary COMPSIZE to size in bytes

Found these while trying to autogenerate a Python API for the Core OpenGL 4.6 commands directly from the XML


Perksey (approved):
Approved. I've expressed my annoyances with COMPSIZE in the past so glad these are being removed where appropriate.

oddhack:
@pdaniell-nv seems unobjectionable to me - this shouldn't affect generated headers. Please mark Signed Off if you agree.

PR #531, opened 2022-07-28
EXT_separate_depth_stencil: Explicit unsupported cases

gnl21:
Adds explicit mention of the conditions that may cause implementations
to return FRAMEBUFFER_UNSUPPORTED for framebuffers with separate depth
and stencil. (The spec currently allows implementations to return
FRAMEBUFFER_UNSUPPORTED for any implementation dependent reason, so this
is not a spec change, but it is more helpful for users of the API if
these conditions are set out in the spec.)


pdaniell-nv (approved):
Looks fine. We can sign off in the upcoming WG meeting unless you need it sooner.

pdaniell-nv:
We approved this change in the previous WG meeting. @oddhack this is approved to merge.

PR #534, opened 2022-08-08
Rename group attribute for non GLenum parameters not corresponding to any enum group

### LINKED ISSUE #517: Group tags that don't correspond to actual enum groups
There is a lot of different parameters that are marked with `group` attributes that are not actually enum groups, but some other kind of metadata. 
Some examples includes:
```xml
<!--from glGetListParameterfvSGIX-->
<param group="CheckedFloat32" len="COMPSIZE(pname)"><ptype>GLfloat</ptype> *<name>params</name></param>

<!--from glGetMultiTexImageEXT-->
<param group="CheckedInt32"><ptype>GLint</ptype> <name>level</name></param>

<!--from glGetTextureImageEXT-->
<param class="texture" group="Texture"><ptype>GLuint</ptype> <name>texture</name></param>
```
These do not denote enum groups like the majority of groups, but instead indicate some other out-of-band information about these parameters.

In total I'm pretty sure that atm there are 1076 functions, with a total of 1849 parameters that are tagged with an group but doesn't have the `GLenum` type.

Unfortunately some of these parameters are actually enums, but their type doesn't reflect this. See #516 for example.
There are 20 functions that fit this description: (written as `<type> <parameter name>: <group>`)
```
glGetDebugMessageLogAMD:
  GLuint* severities: DebugSeverity
glGetFramebufferPixelLocalStorageSizeEXT:
  GLuint target: FramebufferTarget
glMultiTexImage1DEXT:
  GLint internalformat: InternalFormat
glMultiTexImage2DEXT:
  GLint internalformat: InternalFormat
glMultiTexImage3DEXT:
  GLint internalformat: InternalFormat
glQueryObjectParameteruiAMD:
  GLuint param: OcclusionQueryEventMaskAMD
glTexImage1D:
  GLint internalformat: InternalFormat
glTexImage2D:
  GLint internalformat: InternalFormat
glTexImage2DMultisampleCoverageNV:
  GLint internalFormat: InternalFormat
glTexImage3D:
  GLint internalformat: InternalFormat
glTexImage3DMultisampleCoverageNV:
  GLint internalFormat: InternalFormat
glTexStorageAttribs2DEXT:
  GLint* attrib_list: TexStorageAttribs
glTexStorageAttribs3DEXT:
  GLint* attrib_list: TexStorageAttribs
glTextureImage1DEXT:
  GLint internalformat: InternalFormat
glTextureImage2DEXT:
  GLint internalformat: InternalFormat
glTextureImage2DMultisampleCoverageNV:
  GLint internalFormat: InternalFormat
glTextureImage2DMultisampleNV:
  GLint internalFormat: InternalFormat
glTextureImage3DEXT:
  GLint internalformat: InternalFormat
glTextureImage3DMultisampleCoverageNV:
  GLint internalFormat: InternalFormat
glTextureImage3DMultisampleNV:
  GLint internalFormat: InternalFormat
```

There are also a lot of places where parameters of type `GLboolean` are marked with the group `Boolean`, this is technically a valid group as there are entries inside of it, but it would make no sense to change `GLboolean` to `GLenum` (and it would be breaking too).

It would be nice if something could be done about this.

I would like to propose the following course of action.
1. Fix the 20 functions so that they take `GLenum` instead of their current `GLint`/`GLuint` (`uint` might be a problem here).
2. Change the attribute name on the rest of the parameters from `group` to something like `kind`.
3. Now all `group` attributes indicate valid enums, and this way we can work towards complete correctness for enum groups.

This is working towards having robust enum groups that were discussed in #481, this can be used as somewhat of a discussion forum for figuring out how these changes should be done.

Here I'm attaching a file containing a complete breakdown of the functions that either are valid enums but do not sure `GLenum` or have a group attribute that is not a valid enum (with the `Boolean` group filtered out).

[misplaced_groups.txt](https://github.com/KhronosGroup/OpenGL-Registry/files/8928035/misplaced_groups.txt)

NogginBops:
Fixes #517 

I've used the word `kind` here to indicate that the label indicates the kind if value the parameter is (eg `ClampedFloat32` or `CoordF`). 
I was thinking about other words to use such as "attrib", "category", "variety", "set" or "label", but personally it seemed like "kind" was the best, most descriptive name.

@Perksey @SunSerega comments?

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -8313,45 +8313,45 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glClearColor</name></proto>
-            <param kind="Color"><ptype>GLfloat</ptype> <name>red</name></param>
-            <param kind="Color"><ptype>GLfloat</ptype> <name>green</name></param>
-            <param kind="Color"><ptype>GLfloat</ptype> <name>blue</name></param>
-            <param kind="Color"><ptype>GLfloat</ptype> <name>alpha</name></param>
+            <param kind="Clamp01,Color"><ptype>GLfloat</ptype> <name>red</name></param>
+            <param kind="Clamp01,Color"><ptype>GLfloat</ptype> <name>green</name></param>
+            <param kind="Clamp01,Color"><ptype>GLfloat</ptype> <name>blue</name></param>
+            <param kind="Clamp01,Color"><ptype>GLfloat</ptype> <name>alpha</name></param>
             <glx type="render" opcode="130"/>
         </command>
         <command>
             <proto>void <name>glClearColorIiEXT</name></proto>
-            <param><ptype>GLint</ptype> <name>red</name></param>
-            <param><ptype>GLint</ptype> <name>green</name></param>
-            <param><ptype>GLint</ptype> <name>blue</name></param>
-            <param><ptype>GLint</ptype> <name>alpha</name></param>
+            <param kind="Color"><ptype>GLint</ptype> <name>red</name></param>
+            <param kind="Color"><ptype>GLint</ptype> <name>green</name></param>
+            <param kind="Color"><ptype>GLint</ptype> <name>blue</name></param>
+            <param kind="Color"><ptype>GLint</ptype> <name>alpha</name></param>
             <glx type="render" opcode="4292"/>
         </command>
         <command>
             <proto>void <name>glClearColorIuiEXT</name></proto>
-            <param><ptype>GLuint</ptype> <name>red</name></param>
-            <param><ptype>GLuint</ptype> <name>green</name></param>
-            <param><ptype>GLuint</ptype> <name>blue</name></param>
-            <param><ptype>GLuint</ptype> <name>alpha</name></param>
+            <param kind="Color"><ptype>GLuint</ptype> <name>red</name></param>
+            <param kind="Color"><ptype>GLuint</ptype> <name>green</name></param>
+            <param kind="Color"><ptype>GLuint</ptype> <name>blue</name></param>
+            <param kind="Color"><ptype>GLuint</ptype> <name>alpha</name></param>
             <glx type="render" opcode="4293"/>
         </command>
         <command>
             <proto>void <name>glClearColorx</name></proto>
-            <param><ptype>GLfixed</ptype> <name>red</name></param>
-            <param><ptype>GLfixed</ptype> <name>green</name></param>
-            <param><ptype>GLfixed</ptype> <name>blue</name></param>
-            <param><ptype>GLfixed</ptype> <name>alpha</name></param>
+            <param kind="Clamped01,Color"><ptype>GLfixed</ptype> <name>red</name></param>
+            <param kind="Clamped01,Color"><ptype>GLfixed</ptype> <name>green</name></param>
+            <param kind="Clamped01,Color"><ptype>GLfixed</ptype> <name>blue</name></param>
+            <param kind="Clamped01,Color"><ptype>GLfixed</ptype> <name>alpha</name></param>
         </command>
         <command>
             <proto>void <name>glClearColorxOES</name></proto>
-            <param kind="Clamped01"><ptype>GLfixed</ptype> <name>red</name></param>
-            <param kind="Clamped01"><ptype>GLfixed</ptype> <name>green</name></param>
-            <param kind="Clamped01"><ptype>GLfixed</ptype> <name>blue</name></param>
-            <param kind="Clamped01"><ptype>GLfixed</ptype> <name>alpha</name></param>
+            <param kind="Clamped01,Color"><ptype>GLfixed</ptype> <name>red</name></param>
+            <param kind="Clamped01,Color"><ptype>GLfixed</ptype> <name>green</name></param>
+            <param kind="Clamped01,Color"><ptype>GLfixed</ptype> <name>blue</name></param>
+            <param kind="Clamped01,Color"><ptype>GLfixed</ptype> <name>alpha</name></param>
         </command>
         <command>
             <proto>void <name>glClearDepth</name></proto>
-            <param><ptype>GLdouble</ptype> <name>depth</name></param>
+            <param kind="Clamped01"><ptype>GLdouble</ptype> <name>depth</name></param>
             <glx type="render" opcode="132"/>
         </command>
         <command>
@@ -8361,17 +8361,17 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glClearDepthf</name></proto>
-            <param><ptype>GLfloat</ptype> <name>d</name></param>
+            <param kind="Clamped01"><ptype>GLfloat</ptype> <name>d</name></param>
         </command>
         <command>
             <proto>void <name>glClearDepthfOES</name></proto>
-            <param kind="ClampedFloat32"><ptype>GLclampf</ptype> <name>depth</name></param>
+            <param kind="Clamped01"><ptype>GLclampf</ptype> <name>depth</name></param>
             <glx type="render" opcode="4308"/>
             <alias name="glClearDepthf"/>
         </command>
         <command>
             <proto>void <name>glClearDepthx</name></proto>
-            <param><ptype>GLfixed</ptype> <name>depth</name></param>
+            <param kind="Clamped01"><ptype>GLfixed</ptype> <name>depth</name></param>
         </command>
         <command>
             <proto>void <name>glClearDepthxOES</name></proto>
@@ -8621,16 +8621,16 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glColor3fVertex3fSUN</name></proto>
-            <param><ptype>GLfloat</ptype> <name>r</name></param>
-            <param><ptype>GLfloat</ptype> <name>g</name></param>
-            <param><ptype>GLfloat</ptype> <name>b</name></param>
+            <param kind="Color"><ptype>GLfloat</ptype> <name>r</name></param>
+            <param kind="Color"><ptype>GLfloat</ptype> <name>g</name></param>
+            <param kind="Color"><ptype>GLfloat</ptype> <name>b</name></param>
             <param><ptype>GLfloat</ptype> <name>x</name></param>
             <param><ptype>GLfloat</ptype> <name>y</name></param>
             <param><ptype>GLfloat</ptype> <name>z</name></param>
         </command>
         <command>
             <proto>void <name>glColor3fVertex3fvSUN</name></proto>
-            <param len="3">const <ptype>GLfloat</ptype> *<name>c</name></param>
+            <param kind="Color" len="3">const <ptype>GLfloat</ptype> *<name>c</name></param>
             <param len="3">const <ptype>GLfloat</ptype> *<name>v</name></param>
         </command>
         <command>
@@ -8640,14 +8640,14 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glColor3hNV</name></proto>
-            <param><ptype>GLhalfNV</ptype> <name>red</name></param>
-            <param><ptype>GLhalfNV</ptype> <name>green</name></param>
-            <param><ptype>GLhalfNV</ptype> <name>blue</name></param>
+            <param kind="Color"><ptype>GLhalfNV</ptype> <name>red</name></param>
+            <param kind="Color"><ptype>GLhalfNV</ptype> <name>green</name></param>
+            <param kind="Color"><ptype>GLhalfNV</ptype> <name>blue</name></param>
             <vecequiv name="glColor3hvNV"/>
         </command>
         <command>
             <proto>void <name>glColor3hvNV</name></proto>
-            <param len="3">const <ptype>GLhalfNV</ptype> *<name>v</name></param>
+            <param kind="Color" len="3">const <ptype>GLhalfNV</ptype> *<name>v</name></param>
             <glx type="render" opcode="4244"/>
         </command>
         <command>
@@ -8712,13 +8712,13 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glColor3xOES</name></proto>
-            <param><ptype>GLfixed</ptype> <name>red</name></param>
-            <param><ptype>GLfixed</ptype> <name>green</name></param>
-            <param><ptype>GLfixed</ptype> <name>blue</name></param>
+            <param kind="Color"><ptype>GLfixed</ptype> <name>red</name></param>
+            <param kind="Color"><ptype>GLfixed</ptype> <name>green</name></param>
+            <param kind="Color"><ptype>GLfixed</ptype> <name>blue</name></param>
         </command>
         <command>
             <proto>void <name>glColor3xvOES</name></proto>
-            <param len="3">const <ptype>GLfixed</ptype> *<name>components</name></param>
+            <param kind="Color" len="3">const <ptype>GLfixed</ptype> *<name>components</name></param>
         </command>
         <command>
             <proto>void <name>glColor4b</name></proto>
@@ -8756,10 +8756,10 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glColor4fNormal3fVertex3fSUN</name></proto>
-            <param><ptype>GLfloat</ptype> <name>r</name></param>
-            <param><ptype>GLfloat</ptype> <name>g</name></param>
-            <param><ptype>GLfloat</ptype> <name>b</name></param>
-            <param><ptype>GLfloat</ptype> <name>a</name></param>
+            <param kind="Color"><ptype>GLfloat</ptype> <name>r</name></param>
+            <param kind="Color"><ptype>GLfloat</ptype> <name>g</name></param>
+            <param kind="Color"><ptype>GLfloat</ptype> <name>b</name></param>
+            <param kind="Color"><ptype>GLfloat</ptype> <name>a</name></param>
             <param><ptype>GLfloat</ptype> <name>nx</name></param>
             <param><ptype>GLfloat</ptype> <name>ny</name></param>
             <param><ptype>GLfloat</ptype> <name>nz</name></param>
@@ -8769,7 +8769,7 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glColor4fNormal3fVertex3fvSUN</name></proto>
-            <param len="4">const <ptype>GLfloat</ptype> *<name>c</name></param>
+            <param kind="Color" len="4">const <ptype>GLfloat</ptype> *<name>c</name></param>
             <param len="3">const <ptype>GLfloat</ptype> *<name>n</name></param>
             <param len="3">const <ptype>GLfloat</ptype> *<name>v</name></param>
         </command>
@@ -8780,15 +8780,15 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glColor4hNV</name></proto>
-            <param><ptype>GLhalfNV</ptype> <name>red</name></param>
-            <param><ptype>GLhalfNV</ptype> <name>green</name></param>
-            <param><ptype>GLhalfNV</ptype> <name>blue</name></param>
-            <param><ptype>GLhalfNV</ptype> <name>alpha</name></param>
+            <param kind="Color"><ptype>GLhalfNV</ptype> <name>red</name></param>
+            <param kind="Color"><ptype>GLhalfNV</ptype> <name>green</name></param>
+            <param kind="Color"><ptype>GLhalfNV</ptype> <name>blue</name></param>
+            <param kind="Color"><ptype>GLhalfNV</ptype> <name>alpha</name></param>
             <vecequiv name="glColor4hvNV"/>
         </command>
         <command>
             <proto>void <name>glColor4hvNV</name></proto>
-            <param len="4">const <ptype>GLhalfNV</ptype> *<name>v</name></param>
+            <param kind="Color" len="4">const <ptype>GLhalfNV</ptype> *<name>v</name></param>
             <glx type="render" opcode="4245"/>
         </command>
         <command>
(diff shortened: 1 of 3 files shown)


Perksey (changes requested):
Needs to be added to registry.rnc. Otherwise looks good.

SunSerega:
The changes are definitely better than keeping it as is.
But I have my nitpicks, mostly small improvements that are better done now, before the fate of `kind` starts to feel sealed:

- I don't think there is a point in having both `class` and `kind` attributes on the same parameter, like in the case of `texture`, `framebuffer`, etc.
The point of `kind` I can see (besides removing wrong `group` uses) is to indicate some additional info, not necessary in any case, but sometimes good for improving bindings.
When I was adding `class=texture` I looked at all `group=texture`, but didn't delete them, because I didn't want to touch them yet.

- `kind=handleARB` doesn't add any info, because it only exists on parameters of the type `GLhandleARB`.
There are a few cases like the return value of `glGetImageHandleARB`: `GLuint64` is used instead of `GLhandleARB`.
If you consider them, it makes sense to use common `kind` or even `class` to include them.
On the other hand, it seems to me that everything with `GLhandleARB` is outdated. So, I'm not sure if there is a point in adding/keeping `kind`/`class` attributes.

- `Half16NV` is also a copy of the parameter type. But unlike `handleARB` - I don't see the reason to keep it.
Same for `kind=Boolean`. And that one case with `BooleanPointer` (which, if anything, is a `BooleanPointerPointer`).

- There are multiple of `kind=Clamped...`, but they all mean the same. I think it's better to combine them.
Also some of `CheckedFloat32` seems to be more of a "checked and clamped", rather than "checked with errors", as it sounds to me.

- Same for `kind=Color...` and `kind=Coord...`. Type is already defined on the parameter, so it's better to combine to something like `kind=Color` or `kind=ColorValue`. (and separate one for `Coord...`)

- Same for `MaskedColorIndexValue...`, `MaskedStencilValue`, and `StencilValue`.

- On the other hand, floating-point color values (like in `glBlendColor`) are also clamped. `kind="ClampedFloat,ColorValue"`?
Tbh, I expected there to be many more cases where the parameter has multiple kinds at the same time...

- I'm not sure if `cl_context`, `cl_event`, and `vdpauSurfaceNV` should be `kind`'s, `class`'s, or neither...
They definitely have their own namespace. But also maybe they shouldn't be mixed with other classes, which are defined only in- and for OpenGL.

- `DrawBufferName` is not a name, but rather an index (0..15)

- `kind`'s mentioned in the "Extension" table from #428, like `Path`, `FenceNV`, etc. should be classes.
Well, I was planning to make classes for them myself... I can always make another pull after this one.
Note, there are a few of `kind="List"` without `class="display list"`.

- It seems `PathCommand` should instead be a group `PathCoordType`. Though I don't have a thorough understanding of [this](https://registry.khronos.org/OpenGL/extensions/NV/NV_path_rendering.txt), so maybe it's better if someone of NV to confirm.

- `BufferOffset` and `BufferOffsetARB`, as well as `BufferOffset` and `BufferOffsetARB` are the same.

- `PathElement` describes something more general: Type of values in the array is determined by another parameter.
This pattern can be found in many other places, including core spec. And I think I saw a few other `kind`'s with same meaning, but can't find them now...
Regardless, it is not very useful to mark this relation using `kind` attribute, because there is no info on what parameter describes type. So I think this `kind` should be removed, and maybe at some point replaced with different attribute.

- `ProgramCharacterNV`, as far as I can see is not a program specific thing, but rather a regular `char`, disguised as `GLubyte`, just like in `glGetString`, so it should also be `kind=string`. Or maybe change both to `kind=char`, for consistency with other kinds?

- `kind=LineStipple` exists only once and doesn't do anything special. Its parameter can accept any 16-bit number.

- `ReplacementCodeSUN` actually is a group. In #520 I've merged it with `TriangleListSUN`.

---

And a future suggestion, to make `kind` more useful:

- How about `<kind name="..." description="..."/>` somewhere at the top, to make them more well-defined?
Unlike classes, which are solely defined by it's uses and being a namespace, each unique `kind` defines new parameter usage pattern. And that pattern can't always be infered from uses. For instance all the `Clamped...`, as far as I can see, are clamped between 0 and 1.
And the process of making descriptions will probably uncover more things to improve.

oddhack:
Also needs to be added to xml/readme.tex, the schema description document.

pdaniell-nv:
It appears this isn't ready yet for approval. Let me know if you think otherwise.

NogginBops:
Some thoughts I had while working on fixing these.
(will fix the missed color stuff!)

>I'm not sure if cl_context, cl_event, and vdpauSurfaceNV should be kind's, class's, or neither...
They definitely have their own namespace. But also maybe they shouldn't be mixed with other classes, which are defined only in- and for OpenGL.

The `cl_context` and `cl_event` info is available as `struct _cl_context` and `struct _cl_event` so they should not be `kind`s imo.
The same is true for `vdpauSurfaceNV` that is available as `GLvdpauSurfaceNV`.

>`ProgramCharacterNV`, as far as I can see is not a program-specific thing, but rather a regular `char`, disguised as `GLubyte`, just like in `glGetString`, so it should also be `kind=string`. Or maybe change both to `kind=char`, for consistency with other kinds?

Why is `char` more consistent than `String`?

>Same for `MaskedColorIndexValue`..., `MaskedStencilValue`, and `StencilValue`.

I'm not sure what you are referring to in regards to `MaskedStencilValue`, and `StencilValue`.

>How about `<kind name="..." description="..."/>` somewhere at the top, to make them more well-defined?
Unlike classes, which are solely defined by their uses and being a namespace, each unique `kind` defines a new parameter usage pattern. And that pattern can't always be inferred from uses. For instance all the `Clamped...`, as far as I can see, are clamped between 0 and 1.
And the process of making descriptions will probably uncover more things to improve.

I like this idea, though it seems like quite a lot of work.

SunSerega:
> The `cl_context` and `cl_event` info is available as

Agree. But, just a reminder, so far you removed `vdpauSurfaceNV` and kept the other 2.

> Why is `char` more consistent than `String`?

I was sure `kind` is most often relative to type under the pointer. Like here:
```
            <param kind="CheckedInt32" len="COMPSIZE(pname)">const <ptype>GLint</ptype> *<name>params</name></param>
```
It is every referenced value being checked, not the pointer itself.

Though now that I look at it, there are also lots of cases of:
```
            <param kind="Color" len="3">const <ptype>GLbyte</ptype> *<name>v</name></param>
```
Because now you can say it's not `ColorValue` (as I explained in the edit above), it's the whole parameter being a color.

---

On the other hand:
```
            <param kind="CertainKind" len="3"><ptype>ItemType</ptype> ** <name>v</name></param>
```
Note: Double pointer. Is `CertainKind` a kind of `ItemType`? Of the array of `ItemType`? Of the jagged array of `ItemType`?
Well, maybe it would make sense to specify it in the `kind`'s description. That is also fine since every kind is its own pattern.

---

And on the... Third hand - it is important to return from abstractions to actual use.
And I think there (even in the case of the `Color` kind) if the binding is defining a mechanism to handle the kind, that mechanism will probably be applied per value. At least I struggle to find counterexamples.

> I'm not sure what you are referring to in regards to `MaskedStencilValue`, and `StencilValue`.

I remember, from a quick (not thorough) look at spec - I found that `MaskedStencilValue` and `StencilValue` mean the same, i.e. the uses of `StencilValue` are "masked" in all the same way.
But these kinds sure are pretty convoluted and confusing.

---

> like this idea, though it seems like quite a lot of work.

How about this: You add the description to simple kinds, like `String`, `Color`, etc.
And, as for this pull, kinds that are not documented (like `MaskedStencilValue`) are thought of as unsupported (and expected to be heavily changed).
Then unsupported kinds can be incrementally improved. I will also add a warning to the log of my data scrappers for these kinds, to not forget.

---

P.S. You should probably merge the official `main` branch into this one because there are conflicts.

SunSerega:
Also finished updating my initial review, so it's easier to see what's left.

NogginBops:
>Same for `MaskedColorIndexValue`, `MaskedStencilValue`, and `StencilValue`.

`MaskedStencilValue` seems to be used with `GLuint` while `StencilValue` seems to be used for `GLint` types. There probably isn't that much more going on there.

But I'm unsure how `MaskedColorIndexValue` relates to this? It is used in `glIndexMask` and `glClearIndex`. For `glIndexMask` I think the usage is similar, but I don't understand what is going on with `glClearIndex`. In `glIndexMask` it's a proper mask, but in `glClearIndex` it is an index?? (and it's also a `GLfloat`?)

SunSerega:
> but in `glClearIndex` it is an index?? (and it's also a `GLfloat`?)

Yeah, something crazy is happening there. I remember reading about `GLfloat` being converted to fixed, then int, and then also getting masked. At least at the moment, the part where it's masked seemed very similar to `StencilValue`, IIRC because a few lowest bits are taken. But I didn't look into it thoroughly.

NogginBops:
@SunSerega I've made an initial `<kinds>` tag that lists and describes some kinds. The don't need to be final names, and I haven't changed all of the places where they are used. It's more to give a general idea of how the tags could look.

SunSerega:
Looks mostly good so far, updated the review again.

SunSerega:
Trying to apply this to my bindings. In the process found that `SampleMaskNV` also isn't a group. Though, I'm not sure if it should be a kind, because it accepts any bit combination.

SunSerega:
And it's great you also look at unrelated groups, but I think you should fix them in separate PR.
And cases where you removed `ColorPointerType` should probably still have a group, just a different one.

NogginBops:
@SunSerega I've done a, from what I can tell, complete marking of color parameters with `kind=Color`.
During that I found `glClearAccum` that clamps values to the range `[-1, 1]`, I gave it a temporary name but I would like to come up with a better one if possible.

SunSerega:
> I would like to come up with a better one if possible

In the first place, there is no restriction on characters, except `,` (coma).
So how about `Clamped 0..1` and `Clamped -1..+1`?
And maybe then change the separator to `;` (`Clamped 0..1; Color`), for readability.

Perksey:
For list-like things we tend to use commas instead of semicolons. But it’s worth noting this is already too complicated to be maintainable by vendors or Khronos, so given this consider how much augmentation you want to put here vs how much you want to put in your own libraries.

NogginBops:
We don't expect Khronos nor vendors will use or update these. 
As these where groups initially we expect the same policy as with groups, "as long as it doesn't affect the gl.h header and there is a consensus".

And I think these could enable some interesting generation options once we have them complete. And as the specification doesn't change much these days a lot of the work put in now won't need to be monitored and maintained for changes. They will pretty much remain as they are.

NogginBops:
>In the first place, there is no restriction on characters, except , (coma).
So how about Clamped 0..1 and Clamped -1..+1?

Looks funky but I guess we don't need to limit ourselves. An issue with separating numbers with `..` is that parsing floats is harder that way. But putting in floating point number into the clamped kind will likely not happen, and is likely going overboard.

I'm fine with `Clamped -1..1` and `Clamped 0..1`, but it would also be worth to consider `Clamped[0 1]` and `Clamped[-1 1]` as well. What do you think @SunSerega ?

SunSerega:
I don't think float parsing is a concern. It would only make sense if there were many kinds of `Clamped` kind.
However something like `Clamped[-1/+1]` or `Clamped[-1:+1]` would be more readable than `a..b`.
And also looks great with commas as separators: `Clamped[-1:+1], Color`.

Perksey:
> We don't expect Khronos nor vendors will use or update these.

Ok fair enough, figured I'd bring it up given a lot of the design behind the current group representation was to make it as easy as possible for vendors to keep up to date (we've already seen vendors such as ARM and Qualcomm using/updating the groups now it's easy to do so).

pdaniell-nv:
What's the status of this PR? It's not clear to me if there are more changes pending or it's ready for final review and approval.

NogginBops:
There are more changes pending. Will try finish this PR as soon as I have time.

pdaniell-nv:
> There are more changes pending. Will try finish this PR as soon as I have time.

Sure, no worries. I've marked the PR as draft, which you can reset once everything is in. Thanks.

SunSerega:
@NogginBops comma is one symbol that really shouldn't be used in a kind name because it's a name separator.
I would still root for `:` because it better shows the connection between min and max numbers.

NogginBops:
I think it's worth to consider `Clamp[0; 1]` as that would retain a comma, while keeping it a separate character. For some reason I don't want to go for `:` as the separator but if need be we can go for it.

SunSerega:
Also, I think it's a bad idea to add the `SpriteModeSGIX` group to `glSpriteParameterfSGIX` and `glSpriteParameterfvSGIX` (but fine in the other 2 cases you touched).

For high-level languages, it basically says "this should be replaced with a special type", but also it cannot be the main type created for `SpriteModeSGIX`, because it holds values represented in memory as `GLint`.
But things passed to `glSpriteParameterfvSGIX` would need to be stored in another type for `SpriteModeSGIX`, which represents them as floats.

This unnecessarily complicates binding generators.

NogginBops:
> Also, it's a bad idea to add the SpriteModeSGIX group to glSpriteParameterfSGIX and glSpriteParameterfvSGIX (but fine in the other 2 cases you touched).

Why is this a bad idea, the extension document says those are the only accepted values to these functions. I'm fine with removing them though, if the argument is that another extension might add functionality to the function.

SunSerega:
> the extension document says those are the only accepted values to these functions

No, the values from `SpriteModeSGIX` are integers. What `glSpriteParameterfvSGIX` accepts are those values converted to floats. After conversion, they have different representations in memory.

Again, there is no critical problem with that, but there is no point in keeping that information in XML, because while the extension says that only integers can be passed to the last parameter of all 4 (relevant here) functions - only the functions accepting integers are meant to be used. And the other 2 are reserved for the future. Even though they technically already work.

---

Well, to be frank, just not adding group information also doesn't feel like the best solution. But it's a similar case to what `PathElement` was. (look in the "Completed" section of my review)
It's a part of XML that doesn't do any good. Or in the case of `SpriteModeSGIX` on floats - does more harm by polluting with irrelevant info, which is then really awkward to handle in binding generators.
And like `PathElement` - this is a more general case, this time with the `pname` parameter influencing the group of another parameter. If handled properly, it would be much different.

For an example far better then `glSpriteParameter[f,i][,v]SGIX` - look at `glTex[,ture]Parameteri`:
Depending on the `pname` parameter, the `param` can take, besides just a naked integer, the groups:

- `TextureMinFilter`
- `TextureMagFilter`
- `TextureWrapMode`
- `DepthStencilTextureMode`
- `TextureSwizzle`
- `TextureCompareMode`
- `MapTextureFormatINTEL`
- `FoveationConfigBitQCOM`

Most of these groups are defined but unused in XML (they weren't deleted in #520 but caused trouble making that PR).
I'm still in the stage of just forming ideas on how to nicely handle it in XML.

---

And, a bit of step back and look... I said already and I'll say again: I like that you found and made `SpriteModeSGIX` and also added to `glSpriteParameteri[,v]SGIX` functions. But it would be much better as a separate PR.
If you don't want to flood this repo with many small PRs - you can make 1 branch for multiple small unrelated fixes that you find along something big like this.
But, at least for me, it feels really bad to take discussion space of the kind attribute and put here something unrelated. Because the kind attribute is big on itself, there are already many changes and things to talk about.
That may also be the reason it was hard to understand my initial explanation (why `SpriteModeSGIX` on floats bad) - because I tried to make it short. And I still thought but not added some parts/POVs of this new explanation, which might not be as important.

NogginBops:
I can definitely move the `SpriteModeSGIX` enum change out of this PR, I just added it as I was verifying `kind=` attributes. Are there any other such changes that you feel would be better suited for another PR? @SunSerega 
Thinking I'm going to make a kind of "Enum fixes found while making #534" PR.

SunSerega:
Yeah, also note [this](https://github.com/KhronosGroup/OpenGL-Registry/pull/534#issuecomment-1248218027). Every other change not related to the `kind` attribute is trivial, so it doesn't matter in which PR it is.

SunSerega:
> [Changed Clamped[0, 1] to Clamped[0; 1]](https://github.com/KhronosGroup/OpenGL-Registry/pull/534/commits/ed0fc4025e7b8372d925e64d5e27acfe3d532463)

You forgot to rename `Clamped[-1, 1]` into `Clamped[-1; 1]`.

NogginBops:
@SunSerega I removed the `ColorPointerType` changes and made reverted the `SpriteModeSGIX` changes and put them into #546 .

SunSerega:
Finished checking all the PR changes again. Added some info to review and made my bindings generate [this file](https://github.com/SunSerega/POCGL/blob/unstable/DataScraping/XML/GL/Color%20without%20clamp.txt), listing cases parameter should probably be marked as clamped because it's a floating-point color value.
Though I didn't make it check all the floating point types. Only types which were marked as clamped elsewhere:
`GLfixed GLfloat GLdouble GLclampf GLclampd`

NogginBops:
I've gone through the list your have provided. Only one of the functions (`glMinSampleShading`) should have the clamping.
I have provided links to the relevant parts of the spec in the following [file](https://github.com/KhronosGroup/OpenGL-Registry/files/9901581/color.without.clamp.txt).

SunSerega:
Wait, it's weird. #549

NogginBops:
@SunSerega 
> SampleMaskNV also isn't a group. Though it probably shouldn't be a kind, because it accepts any bit combination.

I think the type `GLbitmask` describes this situation exactly, no need for `kind` or `group` tags.

NogginBops:
I have tried updating the `registry.rnc` and `readme.tex` files, but I don't really have any way to verify if I've done so correctly, so would be good if someone could take a look and see if there is anything that needs to be changed.

xml/registry.rnc
```diff
@@ -51,6 +52,17 @@ Type = element type {
...
+# <kinds> defines descriptions for kinds.
+Kinds = element kinds {
+    Kind *
+}
+
+# <kind> defines a description for a single kind.
+Kind = element kind {
+    Name ,
+    attribute desc
+}
```
SunSerega, line 64: I think `Name` should also be marked as an attribute.

Also, it doesn't make sense to me to define `kinds` and `kind` separately. I think this is how you combine them (tho I also didn't run tests):
```suggestion
# <kinds> defines descriptions for kinds
Kinds = element kinds {
    element kind {
        attribute name ,
        attribute desc
    } *
}
```
(also removed the dot at the end of a `#` comment, to make it more uniform with other comments)

NogginBops: `Name` is defined at the end as it's own attribute, I agree that it's somewhat confusing, but it's what most parts of the document do.

I guess I could combine the two definitions, but I tried to follow the structure of the rest of the document.

SunSerega:
I like the idea of `Matrix` kinds: It adds value to XML because it groups 2 kinds of functions, `glProgramUniformMatrix*v` and `glUniformMatrix*v`, describing how their input can be represented as a common struct.
But I think the same idea would be even more useful for vectors. This is a config I have in my bindings:
```
# glUniform[%c:1,2,3,4%][%t:f,d,i,ui%]v
# glUniform[%c:1,2,3,4%][%t:      ui%]vEXT
# glUniform[%c:1,2,3,4%][%t:f,  i   %]vARB
!possible_par_types
 *	| *	| +var Vec{%c%}{%t%}	|
```
(This isn't comprehensive: At the very least, I'm missing all the `glProgramUniform*v`'s)

---

Also, I searched for `UniformMatrix` and there seem to be a few you missing:
- glProgramUniformMatrix2dv
- glProgramUniformMatrix2dvEXT
- glProgramUniformMatrix3x2fvEXT
- glUniformMatrix2dv

NogginBops:
I've added `kind=VectorN` now, it's added for all of the uniform functions but I didn't go through any other functions to add the tag there, but that is also something that could be done.

SunSerega:
Well, there are also `glGetUniform*`, which don't fit (they can be used with any size of vectors)
And there are a few exotic ones: `*Uniformui64v*` and `*UniformHandleui64v*`, which, I think, only accept individual values, not vectors.

But anyway, I think it's better to now fix `Clamp01` and`SampleMaskNV`, and un-draft this pull.
IMO there is nothing left that needed to be done before the first version of the `kind` attribute. The rest can be done in separate incremental pulls.

xml/registry.rnc
SunSerega, line 64: Validated everything with [jing](https://github.com/relaxng/jing-trang), command:
```
jing.jar -c "G:\...\OpenGL-Registry\xml\registry.rnc" "G:\...\OpenGL-Registry\xml\gl.xml"
```
Found only 1 problem, it says the syntax of `desc` is wrong. Works if changed like this:
```suggestion
# <kind> defines a description for a single kind.
Kind = element kind {
    Name ,
    attribute desc { text }
}
```

pdaniell-nv:
@Perksey would you like to give this another review?

xml/gl.xml
```diff
@@ -9072,29 +9110,29 @@ typedef unsigned int GLhandleARB;
             <proto>void <name>glColorTableParameterfv</name></proto>
             <param group="ColorTableTarget"><ptype>GLenum</ptype> <name>target</name></param>
             <param group="ColorTableParameterPName"><ptype>GLenum</ptype> <name>pname</name></param>
-            <param group="CheckedFloat32" len="COMPSIZE(pname)">const <ptype>GLfloat</ptype> *<name>params</name></param>
+            <param kind="CheckedFloat32" len="COMPSIZE(pname)">const <ptype>GLfloat</ptype> *<name>params</name></param>
```
Perksey, line 9113: This kind hasn't been declared at this top of the file.

Perksey, line 9113: I'm fine either way, but wanted to point it out anyway: it seems the precedent set elsewhere in your changes remove type information from the kind names (e.g. `Coord` instead of `CoordF`) which makes sense given you can get this from the `ptype`. Is there any reason why this isn't done with this kind?

Perksey (approved):
I'm pretty happy with these changes, I think the final form of these attributes makes sense given these are meant to be a "hint" and not gospel for what a parameter value represents - it's a good balance.

There are a few things worth double checking (and they may already have been discussed, so apologies if so!) but these needn't block merging, I'm happy with this is as is.

xml/gl.xml
SunSerega: Some `CheckedFloat32` and `CheckedInt32` kinds will need to be turned into `Clamped[*]`, some may be more complex.
I proposed to leave them undocumented to say they are unsupported, while not deleting existing info from XML.
Similar in the case of `Path` and `FenceNV` kinds - but they need to turn into their respective `class` attribute.
Look at the "Long term (probably other PR's)" section of my review.

NogginBops:
I just realized I need to remove the `Clamped[0; 1]` on `glClearColor`, will do that soon.

xml/gl.xml
NogginBops: `CheckedFloat32` is a name that comes from the old groups and I've probably not looked at what kind this particular parameter should have.
I can do a pass over them to see if any of them are easily changed or removed, but otherwise I think we can fix them in future PRs.

NogginBops: I have a working branch where I have already done the changes to remove all of the `CheckedInt32` and `CheckedFloat32` kinds, and removed them or replaced them with the appropriate replacement.
https://github.com/NogginBops/OpenGL-Registry/tree/more-kind-changes

xml/readme.tex
```diff
@@ -247,6 +251,62 @@ \subsection{Example of a \tag{types} tag}
...
+\label{tag:kind}
+
+Each \{kind} tag defines a single kind annotation.
+
+\subsection{Attributes of \tag{kind} tag}
+
+\begin{itemize}
+\item \attr{name} - kind name, an arbitrary string used
+      to identify this kind.
+\item \attr{desc} - kind description, a description of what it
```
oddhack, line 292: Since ',' is the separator for multiple kind annotations, the kind 'desc' presumably cannot contain ',' - this should be mentioned here.

oddhack (changes requested):
Aside from the note on restricting kind annotations to not contain the ',' separator, I'll need to run this through the schema validator and ensure the generated headers aren't affected before merging, but it seems fine.

xml/readme.tex
SunSerega: `desc` is a description of one kind, it can contain anything. Did you mean the `name` attribute?

NogginBops:
Should the `kind="Color"` attribute be changed to `kinds="Color"` to indicate that it is a list of kinds instead of a single kind.
I initially thought about `kind=""` as a single value, and it is kind of stuck in my mind, and I'm wondering if changing the plurality of the word used would be more clear?

xml/readme.tex
NogginBops: I think I have resolved this now.

Perksey:
I think it's fine as is, as to me it reads as "the parameter is of a kind that is a color and is clamped to a specific range"

pdaniell-nv:
@NogginBops could you resolve the merge conflicts? I assume otherwise this is ready to go in? Thanks.

pdaniell-nv:
> Aside from the note on restricting kind annotations to not contain the ',' separator, I'll need to run this through the schema validator and ensure the generated headers aren't affected before merging, but it seems fine.

@oddhack did you get a chance to run this through the schema validator?

pdaniell-nv:
@oddhack, this is approved to merge as soon as the merge conflicts are resolved and you're happy with it.

NogginBops:
Fixed merge conflicts.

PR #537, opened 2022-08-25
Add some enum groups for ANGLE use cases.

null77:
This includes a few GLES 1.x enumes and a few others that we use
in ANGLE.

See http://anglebug.com/6461 for context.

@pdaniell-nv @oddhack PTAL


null77:
Hm, looks like the version of this repo that we have mirrored is very out of date:

https://chromium.googlesource.com/external/github.com/KhronosGroup/OpenGL-Registry/

currently on rev 5048bd8682a4483d00ea8de495cd8e59b2c720af . I'll see if I can get that fixed and then resubmit.

PR #540, opened 2022-09-07
Clarify that PLS doesn't allow leak from outside framebuffer

ShabbyX:
Per internal issue #168


pdaniell-nv:

Issue #532, opened 2022-08-04
Blit layered framebuffers


bl4ckb0ne:
I'm trying to copy a cubemap texture between 2 framebuffers using blits. Both 
framebuffers are complete layered framebuffers with 6 layers each, with the
same size.

When doing a simple blit operation, as seen below, only one face is copied.

```c
glBindFramebuffer(GL_DRAW_FRAMEBUFFER, dst);
glClearColor(1.f, 1.f, 1.f, 1.f);
glClear(GL_COLOR_BUFFER_BIT);

glBindFramebuffer(GL_READ_FRAMEBUFFER, dst);
glBlitFramebuffer(0, 0, width, height, 0, 0, width, height, GL_COLOR_BUFFER_BIT, GL_NEAREST);
```

According to the GLES 3.2 spec [1] 9.8 Layered framebuffers

> When  commands  such  as ReadPixels read  from  a  layered  framebuffer,  theimage at layer zero of the selected attachment is always used to obtain pixel values

But the `glClear` call before the blit behaves differently

> When theClearorClearBuffer*commands described in section 15.2.3 areused to clear a layered framebuffer attachment,  all layers of the attachment arecleared.

Why is the blit operation only possible on the first layer but the clear is
possible on all layers? I there another way to achieve what I want to do? Could
an extension fix this?

I'd be happy to draft something if needed to fix this issue.

[1]: https://registry.khronos.org/OpenGL/specs/es/3.2/es_spec_3.2.pdf

pdaniell-nv:
On page 540 of the [OpenGL 4.6 core specification](https://registry.khronos.org/OpenGL/specs/gl/glspec46.core.pdf) when describing glBlitFramebuffer it says:
> If the read framebuffer is layered (see section 9.8), pixel values are read from
layer zero. If the draw framebuffer is layered, pixel values are written to layer zero.
If both read and draw framebuffers are layered, the blit operation is still performed
only on layer zero.

To blit non-zero layers you can attach the texture to the framebuffer with glFramebufferTexture3D() or glFramebufferTextureLayer() and specify the layer parameter as non-zero.

bl4ckb0ne:
I have the following piece of code to achieve what I want but it feels heavy

```c

glBindFramebuffer(GL_DRAW_FRAMEBUFFER, dst);
glBindFramebuffer(GL_READ_FRAMEBUFFER, src);

for (int i = 0; i < 6; ++i) {
[...]

PR #541, opened 2022-09-16
Add GL_EXT_framebuffer_blit_layers

bl4ckb0ne:
Closes https://github.com/KhronosGroup/OpenGL-Registry/issues/532

Mesa implementation https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/18847

### DIFF
--- a/extensions/EXT/EXT_framebuffer_blit_layers.txt
+++ b/extensions/EXT/EXT_framebuffer_blit_layers.txt
@@ -1,0 +1,111 @@
+Name
+
+    EXT_framebuffer_blit_layers
+
+Name Strings
+
+    GL_EXT_framebuffer_blit_layers
+
+Contacts
+
+    Simon Zeni, Status Holdings Ltd. <simon@zeni.ca>
+
+Contributors
+
+    Simon Zeni
+
+Status
+
+    Draft
+
+Version
+
+    Version 0.1, 2022/09/16
+
+Number
+
+    OpenGL Extension #558
+    OpenGL ES Extension #341
+
+Dependencies
+
+    OpenGL 3.0, OpenGL ES 3.0, or GL_EXT_framebuffer_blit required
+
+    This extension is written against the OpenGL 4.6 Core and OpenGL ES 3.2
+    specifications.
+
+Overview
+
+    This extensions defines the behaviour for copying data from one layered
+    framebuffer to another layered framebuffer.
+
+New Procedures and Functions
+
+    void BlitFramebufferLayersEXT(int srcX0, int srcY0, int srcX1, int srcY1,
+                                  int dstX0, int dstY0, int dstX1, int dstY1,
+                                  bitfield mask, enum filter);
+
+    void BlitFramebufferLayerEXT(int srcX0, int srcY0, int srcX1, int srcY1,
+                                 int srcLayer, int dstX0, int dstY0, int dstX1,
+                                 int dstY1, int dstLayer, bitfield mask,
+                                 enum filter);
+
+New Tokens
+
+    None
+
+Additions to Chapter 16 of the OpenGL ES 3.2 specification (Reading and
+Copying Pixels) and Chaper 18 of the OpenGL 4.6 Core specification (Reading
+and Copying Pixels)
+
+    Append the following to the section 16.2.1 (Blitting Pixel Rectangles) of
+    the OpenGL ES 3.2 specification, and to the section 18.3.1 (Blitting Pixel
+    Rectangles) of the OpenGL 4.6 core specification.
+
+    The command
+
+        void BlitFramebufferLayersEXT(int srcX0, int srcY0, int srcX1,
+            int srcY1, int dstX0, int dstY0, int dstX1, int dstY1,
+            bitfield mask, enum filter);
+
+    extends the behaviour of the command BlitFramebuffer to all the layers of
+    the textures bound to the read framebuffer and the draw framebuffer.
+
+    Calling BlitFramebufferLayersEXT will result in an INVALID_OPERATION if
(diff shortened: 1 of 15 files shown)


extensions/EXT/EXT_framebuffer_blit_layers.txt
```diff
@@ -0,0 +1,95 @@
...
+    draw framebuffer.
+
+    Calling BlitFramebufferLayerEXT will result in an INVALID_OPERATION if
+    srcLayer or dstLayer is negative, larger than the value of 
+    MAX_ARRAY_TEXTURE_LAYERS minus one, or larger than the number of layers in
+    the attached texture.
+
+Issues
+
+    None
```
janharaldfredriksen-arm, line 90: GL_OVR_multiview defines an error condition:
--
    INVALID_FRAMEBUFFER_OPERATION is generated by commands that read from the
    framebuffer such as BlitFramebuffer, ReadPixels, CopyTexImage*, and
    CopyTexSubImage*, if the number of views in the current read framebuffer
    is greater than 1.
--
That should probably apply to BlitFramebufferLayer*EXT too?

bl4ckb0ne: Good catch, I'll add it. FWIW OVR_multiview2 [1] relaxes the restrictions.

[1]: https://registry.khronos.org/OpenGL/extensions/OVR/OVR_multiview2.txt

bl4ckb0ne: After reading the extension, I think `GL_OVR_multiview` refers to views as eye views, similar to what OpenXR does [1].

> The most obvious use case in this model is to support two simultaneous views: one view for each eye.

I don't think that counts framebuffer layers.

[1]: https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#view_configurations

bl4ckb0ne: > Finally, multiple layers of two-dimensional array textures can be
    attached to an attachment point. Such attachments represent multiple
    views, and the corresponding attachment point is considered to be
    _multiview_.

Nevermind, Ill add it.

bl4ckb0ne:
MESA implementation is done, waiting for review. This is also done for review.

ping @oddhack @pdaniell-nv @cubanismo

extensions/EXT/EXT_framebuffer_blit_layers.txt
```diff
@@ -0,0 +1,107 @@
...
+Number
+
+    OpenGL Extension #558
+    OpenGL ES Extension #341
+
+Dependencies
+
+    OpenGL ES 3.0 is required
+
+    This extension is written against the OpenGL ES 3.2 specification
```
pdaniell-nv, line 34: The edits below are against the OpenGL ES 3.0 extension, which conflicts with this statement. Do the below edits need to be moved to OpenGL ES 3.2?

pdaniell-nv: Also since this extension is for OpenGL too it should say which version of the OpenGL extension it affects.

extensions/EXT/EXT_framebuffer_blit_layers.txt
```diff
@@ -0,0 +1,107 @@
...
+    void BlitFramebufferLayerEXT(int srcX0, int srcY0, int srcX1, int srcY1,
+                                 int srcLayer, int dstX0, int dstY0, int dstX1,
+                                 int dstY1, int dstLayer, bitfield mask,
+                                 enum filter);
+
+New Tokens
+
+    None
+
+Additions to Chapter 4 of the OpenGL 3.0 Specification (Per Fragment Operations
```
pdaniell-nv, line 56: This should also mention where in the OpenGL spec this addition is made. I guess OpenGL 4.6 section 18.3.1.

cubanismo:
Besides Piers' comments, looks good to me.

extensions/EXT/EXT_framebuffer_blit_layers.txt
bl4ckb0ne: Shouldnt it be OpenGL 3? It introduces `glBlitFramebuffer`, same as ES 3.0

pdaniell-nv:
@bl4ckb0ne are you planning to make an update based on the feedback? It should be simple. Thanks.

extensions/EXT/EXT_framebuffer_blit_layers.txt
bl4ckb0ne: I don't think it needs anything specific from ES 3.2, I'll move it back to ES 3.0.

bl4ckb0ne:
Yes! I forgot to send the comment yesterday, I'm working on integrating your feedback. I just had one question regarding writing the spec against GL 3.0 (which introduces `glBlitFramebuffer`) or 4.6

extensions/EXT/EXT_framebuffer_blit_layers.txt
pdaniell-nv: The extension can apply edits to any version of the core specs, it's whatever version is mentioned in the Dependencies section. In general extensions usually apply edits to the latest versions of the core specs. The functionality can still apply to earlier versions. So in this case the Dependencies section might say:
```
Dependencies

OpenGL 3.0, OpenGL ES 3.0, or GL_EXT_framebuffer_blit required

This extension is written against the OpenGL 4.6 Core and OpenGL ES 3.2 specifications.
```

bl4ckb0ne: Thank you for the answer, I'll finish the edits today

PR #544, opened 2022-10-05
SSO no longer requires redeclaration of gl_Position

ShabbyX:
This behavior was not adopted by core GLSL, and is inconsistently implemented by drivers.  CTS does not follow this requirement, and no known apps exist that do this.

Per internal issue [#169](https://gitlab.khronos.org/opengl/API/-/issues/169)


ShabbyX:
Turns out there are existing drivers that enforce this. Although that's done by failing link rather than compilation.

Issue #545, opened 2022-10-08
glCreateShaderProgramvEXT "strings" parameter does not match glCreateShaderProgramv


null77:
From `gl.xml`:

```
        <command>
            <proto class="program"><ptype>GLuint</ptype> <name>glCreateShaderProgramv</name></proto>
            <param group="ShaderType"><ptype>GLenum</ptype> <name>type</name></param>
            <param><ptype>GLsizei</ptype> <name>count</name></param>
            <param len="count">const <ptype>GLchar</ptype> *const*<name>strings</name></param>
        </command>
        <command>
            <proto class="program"><ptype>GLuint</ptype> <name>glCreateShaderProgramvEXT</name></proto>
            <param group="ShaderType"><ptype>GLenum</ptype> <name>type</name></param>
            <param><ptype>GLsizei</ptype> <name>count</name></param>
            <param len="count">const <ptype>GLchar</ptype> **<name>strings</name></param>
        </command>
```

Notably `strings` has conflicting types `const GLchar *const*` and `const GLchar **`. These are not compatible types in C-based languages so it would be good to resolve the type mismatch. Most likely the EXT has the incorrect annotation.

Perksey:
Is it possible to resolve this issue without breaking users of either function?

NogginBops:
It does look like the specification and `EXT_separate_shader_objects` have different definitions of this function.

4.6 spec:
```c
uint CreateShaderProgramv( enum type, sizei count, const char * const *strings );
```

EXT_separate_shader_objects:
```c
uint CreateShaderProgramEXT(enum type, const char *string);
```

not sure when and where the difference appeared but it's where.

Making a change where would at least not be ABI breaking (at least to my knowledge). 
Is it a problem that they have different definitions? 


My initial thought would be to leave things as they are as the two functions are ABI compatible, and leaving the assumptions the same as when driver support for the extension was being developed reduces the risk of some driver doing some weird modification even though the argument is specified as `const`.
Though if there is a good solution or explanation that would be more optimal for sure.

SunSerega:
>These are not compatible types in C-based languages

They are, but only as a one-way conversion:
```
	const char** s1 = nullptr;
	// OK: Only adding const
	const char* const* s2 = s1;
	// Error: Cannot implicitly remove const
	const char** s3 = s2;
```
[...]

PR #548, opened 2022-10-31
[XML] Made `glCreateShaderProgramvEXT` same as `glCreateShaderProgramv`

SunSerega:
Maybe we should also touch [this](https://github.com/KhronosGroup/OpenGL-Registry/blob/eae1d6dde1e283f6fdf803274a2484007e592599/extensions/EXT/EXT_separate_shader_objects.gles.txt) since related function got changed?

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -10302,7 +10302,7 @@ typedef unsigned int GLhandleARB;
             <proto class="program"><ptype>GLuint</ptype> <name>glCreateShaderProgramvEXT</name></proto>
             <param group="ShaderType"><ptype>GLenum</ptype> <name>type</name></param>
             <param><ptype>GLsizei</ptype> <name>count</name></param>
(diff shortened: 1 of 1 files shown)


NogginBops:
Link to the original issue for completeness: #545

pdaniell-nv:

null77:
@NogginBops I think you forgot to regenerate the headers after this change. It means the extension header and the XML are in conflict as to the type of this function.

null77:
> @NogginBops I think you forgot to regenerate the headers after this change. It means the extension header and the XML are in conflict as to the type of this function.

Sorry, meant to attn @SunSerega

NogginBops:
I think @oddhack might be able to fix this more quickly?

SunSerega:
I think it was explicitly asked not to include header changes in the PR, to avoid conflicts between PRs.
And the headers here are just one version of OpenGL bindings. So like any other bindings, it can be a bit behind the XML.

PR #550, opened 2022-11-06
[xml] Fix param group in DrawElements* commands

hbirchtree:
Ensure that the DrawElements* commands use the proper DrawElementsType
group for type params. Some of these were set to PrimitiveType


NogginBops:
I have no objections.

hbirchtree:
I should note I don't have write permissions, so feel free to merge if the changes are acceptable :)

pdaniell-nv:

PR #551, opened 2022-11-06
[xml] Add missing len attribute to data parameters

hbirchtree:


NogginBops:
Looks good to me.

pdaniell-nv:

PR #552, opened 2022-11-28
SSO no longer requires redeclaration of gl_Position

ShabbyX:
This behavior was not adopted by core GLSL, and is inconsistently implemented by drivers.  CTS does not follow this requirement, and no known apps exist that do this.

Per internal issue #169


ShabbyX:
There's one little fix actually needed

ShabbyX:
Nevermind, there are drivers that do check this, although they don't fail compilation, just link.

PR #546, opened 2022-10-26
Enum fixes found while making #534

NogginBops:
These are some enum fixes that I found while making #534.

PR #560, opened 2023-01-16
[XML] Remove `TextureEnvParameter` enum group dup

SunSerega:
These were added in #546. Only noticed now, because it doesn't affect anything.


pdaniell-nv (approved):

pdaniell-nv:
@oddhack this is approved if you're happy with it.

PR #569, opened 2023-03-24
[WIP] Added len and group attributes to wgl.xml

NogginBops:
This PR aims to add some of the metadata present in `gl.xml` into `wgl.xml`.

One problem I've found is that if needing to reference enums or enum groups from `gl.xml`, and I do think the namespaces should be kept separate. Maybe the GL enum values could be added under some tag in the xml? Or prefix the tags with something like `GL::TextureTarget` for example. None of the solutions stand out as the "correct" one.

### DIFF
--- a/xml/wgl.xml
+++ b/xml/wgl.xml
@@ -72,44 +72,44 @@ Registry at https://github.com/KhronosGroup/OpenGL-Registry
          sometimes reused for other purposes -->
 
     <enums namespace="WGLLayerPlaneMask" type="bitmask" vendor="MS">
-        <enum value="0x00000001"  name="WGL_SWAP_MAIN_PLANE"/>
-        <enum value="0x00000002"  name="WGL_SWAP_OVERLAY1"/>
-        <enum value="0x00000004"  name="WGL_SWAP_OVERLAY2"/>
-        <enum value="0x00000008"  name="WGL_SWAP_OVERLAY3"/>
-        <enum value="0x00000010"  name="WGL_SWAP_OVERLAY4"/>
-        <enum value="0x00000020"  name="WGL_SWAP_OVERLAY5"/>
-        <enum value="0x00000040"  name="WGL_SWAP_OVERLAY6"/>
-        <enum value="0x00000080"  name="WGL_SWAP_OVERLAY7"/>
-        <enum value="0x00000100"  name="WGL_SWAP_OVERLAY8"/>
-        <enum value="0x00000200"  name="WGL_SWAP_OVERLAY9"/>
-        <enum value="0x00000400"  name="WGL_SWAP_OVERLAY10"/>
-        <enum value="0x00000800"  name="WGL_SWAP_OVERLAY11"/>
-        <enum value="0x00001000"  name="WGL_SWAP_OVERLAY12"/>
-        <enum value="0x00002000"  name="WGL_SWAP_OVERLAY13"/>
-        <enum value="0x00004000"  name="WGL_SWAP_OVERLAY14"/>
-        <enum value="0x00008000"  name="WGL_SWAP_OVERLAY15"/>
-        <enum value="0x00010000"  name="WGL_SWAP_UNDERLAY1"/>
-        <enum value="0x00020000"  name="WGL_SWAP_UNDERLAY2"/>
-        <enum value="0x00040000"  name="WGL_SWAP_UNDERLAY3"/>
-        <enum value="0x00080000"  name="WGL_SWAP_UNDERLAY4"/>
-        <enum value="0x00100000"  name="WGL_SWAP_UNDERLAY5"/>
-        <enum value="0x00200000"  name="WGL_SWAP_UNDERLAY6"/>
-        <enum value="0x00400000"  name="WGL_SWAP_UNDERLAY7"/>
-        <enum value="0x00800000"  name="WGL_SWAP_UNDERLAY8"/>
-        <enum value="0x01000000"  name="WGL_SWAP_UNDERLAY9"/>
-        <enum value="0x02000000"  name="WGL_SWAP_UNDERLAY10"/>
-        <enum value="0x04000000"  name="WGL_SWAP_UNDERLAY11"/>
-        <enum value="0x08000000"  name="WGL_SWAP_UNDERLAY12"/>
-        <enum value="0x10000000"  name="WGL_SWAP_UNDERLAY13"/>
-        <enum value="0x20000000"  name="WGL_SWAP_UNDERLAY14"/>
-        <enum value="0x40000000"  name="WGL_SWAP_UNDERLAY15"/>
+        <enum value="0x00000001"  name="WGL_SWAP_MAIN_PLANE" group="Plane"/>
+        <enum value="0x00000002"  name="WGL_SWAP_OVERLAY1" group="Plane"/>
+        <enum value="0x00000004"  name="WGL_SWAP_OVERLAY2" group="Plane"/>
+        <enum value="0x00000008"  name="WGL_SWAP_OVERLAY3" group="Plane"/>
+        <enum value="0x00000010"  name="WGL_SWAP_OVERLAY4" group="Plane"/>
+        <enum value="0x00000020"  name="WGL_SWAP_OVERLAY5" group="Plane"/>
+        <enum value="0x00000040"  name="WGL_SWAP_OVERLAY6" group="Plane"/>
+        <enum value="0x00000080"  name="WGL_SWAP_OVERLAY7" group="Plane"/>
+        <enum value="0x00000100"  name="WGL_SWAP_OVERLAY8" group="Plane"/>
+        <enum value="0x00000200"  name="WGL_SWAP_OVERLAY9" group="Plane"/>
+        <enum value="0x00000400"  name="WGL_SWAP_OVERLAY10" group="Plane"/>
+        <enum value="0x00000800"  name="WGL_SWAP_OVERLAY11" group="Plane"/>
+        <enum value="0x00001000"  name="WGL_SWAP_OVERLAY12" group="Plane"/>
+        <enum value="0x00002000"  name="WGL_SWAP_OVERLAY13" group="Plane"/>
+        <enum value="0x00004000"  name="WGL_SWAP_OVERLAY14" group="Plane"/>
+        <enum value="0x00008000"  name="WGL_SWAP_OVERLAY15" group="Plane"/>
+        <enum value="0x00010000"  name="WGL_SWAP_UNDERLAY1" group="Plane"/>
+        <enum value="0x00020000"  name="WGL_SWAP_UNDERLAY2" group="Plane"/>
+        <enum value="0x00040000"  name="WGL_SWAP_UNDERLAY3" group="Plane"/>
+        <enum value="0x00080000"  name="WGL_SWAP_UNDERLAY4" group="Plane"/>
+        <enum value="0x00100000"  name="WGL_SWAP_UNDERLAY5" group="Plane"/>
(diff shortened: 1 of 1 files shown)


NogginBops:
@SunSerega @Perksey might be interested to comment on this PR.

SunSerega:
`ColorBuffer` and `PixelType` are already found in `gl.xml` and are separate from groups of the same name you added.

xml/wgl.xml
```diff
@@ -72,44 +72,44 @@ Registry at https://github.com/KhronosGroup/OpenGL-Registry
...
+        <enum value="0x01000000"  name="WGL_SWAP_UNDERLAY9" group="Plane"/>
+        <enum value="0x02000000"  name="WGL_SWAP_UNDERLAY10" group="Plane"/>
+        <enum value="0x04000000"  name="WGL_SWAP_UNDERLAY11" group="Plane"/>
+        <enum value="0x08000000"  name="WGL_SWAP_UNDERLAY12" group="Plane"/>
+        <enum value="0x10000000"  name="WGL_SWAP_UNDERLAY13" group="Plane"/>
+        <enum value="0x20000000"  name="WGL_SWAP_UNDERLAY14" group="Plane"/>
+        <enum value="0x40000000"  name="WGL_SWAP_UNDERLAY15" group="Plane"/>
     </enums>
 
     <enums namespace="WGLColorBufferMask" type="bitmask" vendor="ARB">
```
Perksey, line 108: It looks like a pretty decent attempt has been made when originally writing the XML spec to group up each enum into namespaces. Can we have the group attributes make use of these existing namespaces? (it's better if we reuse as much ahead-of-time bikeshedding of the names as possible, given these names are already approved)

xml/wgl.xml
```diff
@@ -72,44 +72,44 @@ Registry at https://github.com/KhronosGroup/OpenGL-Registry
          sometimes reused for other purposes -->
```
Perksey, line 73: Note this comment (was originally in gl.xml but must've got removed):
```xml

    <!-- SECTION: GL parameter class type definitions.
         The groups are intended to contain all the possible legal values
         for corresponding function parameters, but it is likely that many
         of the groups are out of date relative to current OpenGL and OpenGL
         ES specifications, and the many extensions to those specifications.
         As such, they may not be a reliable source for enumeration info.

         We welcome assistance from the community in achieving and
         maintaining the completeness of the enum groups. Khronos does not
         use the enum group information, and the OpenGL Working Group does
         not have internal resources to bring it up to date.
    -->
```

This was added in #280 and #291. While thanks to community efforts this situation has got a lot better, I do think it is worth adding a comment in both files that the groups at the very least **are exclusively community-supported, and that the OpenGL Working Group has no intention of keeping the groupings up-to-date themselves due to them not being relevant for Khronos uses**. It's annoying we missed this comment being removed from gl.xml.

xml/wgl.xml
```diff
@@ -413,13 +413,13 @@ Registry at https://github.com/KhronosGroup/OpenGL-Registry
...
-        <enum value="0x21A4"        name="WGL_GPU_CLOCK_AMD"/>
-        <enum value="0x21A5"        name="WGL_GPU_NUM_PIPES_AMD"/>
+        <enum value="0x21A3"        name="WGL_GPU_RAM_AMD" group="GPUProperty"/>
+        <enum value="0x21A4"        name="WGL_GPU_CLOCK_AMD" group="GPUProperty"/>
+        <enum value="0x21A5"        name="WGL_GPU_NUM_PIPES_AMD" group="GPUProperty"/>
         <enum value="0x21A5"        name="WGL_TEXTURE_RECTANGLE_ATI" comment="Duplicates unrelated WGL_GPU_NUM_PIPES_AMD"/>
-        <enum value="0x21A6"        name="WGL_GPU_NUM_SIMD_AMD"/>
-        <enum value="0x21A7"        name="WGL_GPU_NUM_RB_AMD"/>
-        <enum value="0x21A8"        name="WGL_GPU_NUM_SPI_AMD"/>
+        <enum value="0x21A6"        name="WGL_GPU_NUM_SIMD_AMD" group="GPUProperty"/>
```
Perksey, line 420: What's our ruling on vendor-specific enum groups? In gl.xml this would've likely been `GPUPropertyAMD`. Related: #481

SunSerega:
> Yeah, when making this PR I was envisioning GL, WGL and GLX as different enum group namespaces

Hm... Is it possible for, for instance, the `WGL` function to accept/return an enum from gl.xml?
If it ever becomes needed in the future, how could it be denoted after separating groups into namespaces based on XML files?

NogginBops:
> > Yeah, when making this PR I was envisioning GL, WGL and GLX as different enum group namespaces
> 
> Hm... Is it possible for, for instance, a `WGL` function to accept/return an enum from gl.xml? If it ever becomes needed in the future, how could it be denoted after separating groups into namespaces based on XML files?

Yes, this happens and is marked with a few FIXMEs comments in the PR. I guess this can be solved by making them the same namespace and add a WGL prefix to all groups defined in `wgl.xml`.

To me it seemed more cleaner to have them be different namespaces and figure out a way to cross reference, rather than make them the same namespace but add a prefix to the names in wgl and glx. I initially thought of it like this as it fit better into the current structure of our generator, but it's fine to change that.

There are some downsides to sharing the namespace between files, group changes in `gl.xml` would then have to be checked against `wgl.xml` and `glx.xml` which could very easily be forgotten, vice versa for changes in `wgl.xml` and `glx.xml`.

I think there is also another problem where some WGL functions only accept a set of enums that do not have groups defined in `gl.xml`, so these would have to be added to `gl.xml` and look like an unused group, but in fact it's used by `wgl.xml` and potentially `glx.xml`.

xml/wgl.xml
NogginBops: I'm fine with making this enum vendor specific, but I think most enum groups in `gl.xml` do not have a vendor postfix, even if it's mostly an extension concept. For this group, if someone else made an extension it would likely not fit into the same enum groups so it might reduce future confusion (but also, gl, wgl, and glx aren't getting that many new changes to them and probably won't).

Then there is also the fact that almost all API in wgl comes in the form of extensions, so basically all groups would need a postfix. But maybe that is what we want.

xml/wgl.xml
NogginBops: We can add this comment to these files too.

xml/wgl.xml
NogginBops: Look at that, totally missed these when doing the PR... 😅

SunSerega:
Not for this PR but...
```
        <command>
            <proto><ptype>__GLXextFuncPtr</ptype> <name>glXGetProcAddressARB</name></proto>
            <param>const <ptype>GLubyte</ptype> *<name>procName</name></param>
        </command>
        <command>
            <proto><ptype>__GLXextFuncPtr</ptype> <name>glXGetProcAddress</name></proto>
            <param>const <ptype>GLubyte</ptype> *<name>procName</name></param>
        </command>
```
I guess think would also need to be `kind="gl::String"`?
Feels awkward to dup `kind` definitions between the XML files...

NogginBops:
> Not for this PR but...
> 
> ```
>         <command>
>             <proto><ptype>__GLXextFuncPtr</ptype> <name>glXGetProcAddressARB</name></proto>
>             <param>const <ptype>GLubyte</ptype> *<name>procName</name></param>
>         </command>
>         <command>
>             <proto><ptype>__GLXextFuncPtr</ptype> <name>glXGetProcAddress</name></proto>
>             <param>const <ptype>GLubyte</ptype> *<name>procName</name></param>
>         </command>
> ```
> 
> I guess think would also need to be `kind="gl::String"`? Feels awkward to dup `kind` definitions between the XML files...

I agree these should be marked with the `kind` attribute.
And I'm guessing you are ok with using a prefix (`gl::`, `wgl::`, and `glx::`) to reference between the xml files?
I could put that into the xml spec for the group and kind tags.

SunSerega:
Yes, I've since rewritten my data scrappers. Had too much legacy code, your idea with group namespaces was the final push.

Right now I'm using XML files from the branch in my own fork, which has the changes from this PR plus a bunch of my own, with the namespace prefixes supported for all sorts of items, including the kinds too.

NogginBops:
Ok great then I will move forward on this assuming we are going to go with namespaces.
Will document how the namespaces work in the xml specification before marking this as ready for review.

NogginBops:
@SunSerega how do you feel about doing something like this in `gl.xml`:
```xml
<enum value="0x0DE1" name="GL_TEXTURE_2D" group="...,wgl::ObjectType"/>
<enum value="0x0DE1" name="GL_TEXTURE_3D" group="...,wgl::ObjectType"/>
```
to be able to create enum groups for `wgl.xml`? or should I create a group called `WGLObjectType` (or something like `WGLObjectTypeDX`)

This relates to this to the DX interop extension but will be applicable to other extensions as well
https://registry.khronos.org/OpenGL/extensions/NV/WGL_NV_DX_interop.txt

SunSerega:
Yeah, this also works with how I imagined it: If no namespace is specified - it defaults to the same namespace as the file name. Otherwise, any namespace can be specified.

NogginBops:
@SunSerega what do you think about my newest commit? Is it in the right direction?
It's quite small, I'm just trying to figure out how we want to structure this.

xml/wgl.xml
```diff
@@ -736,8 +734,7 @@ Registry at https://github.com/KhronosGroup/OpenGL-Registry
             <param><ptype>HANDLE</ptype> <name>hDevice</name></param>
             <param len="1">void *<name>dxObject</name></param>
             <param><ptype>GLuint</ptype> <name>name</name></param>
-            <!-- FIXME: reference GL enums? -->
-            <param ><ptype>GLenum</ptype> <name>type</name></param>
+            <param group="gl::WGLObjectTypeDX"><ptype>GLenum</ptype> <name>type</name></param>
```
SunSerega, line 737: `wgl::ObjectTypeDX`?

NogginBops: Ah yeah missed this one, I originally created the enum group in `gl::` but changed it later. Will fix

SunSerega:
Since we are going for file-based namespaces, besides `WGLObjectTypeDX`, there are a few more groups that start with redundant `WGL`:
```
wgl::WGLColorBufferMask
wgl::WGLContextFlagsMask
wgl::WGLContextProfileMask
wgl::WGLDXInteropMask
wgl::WGLImageBufferMask
wgl::WGLLayerPlaneMask
```

NogginBops:
@SunSerega is this PR getting ready or is there something else needed before I can un-draft it?

SunSerega (approved):
Checked everything again, LGTM.

xml/wgl.xml
```diff
@@ -766,6 +766,7 @@ Registry at https://github.com/KhronosGroup/OpenGL-Registry
         <command>
             <proto><ptype>UINT</ptype> <name>wglEnumerateVideoCaptureDevicesNV</name></proto>
             <param><ptype>HDC</ptype> <name>hDc</name></param>
+            <!-- FIXME: Some way for len="" to refer to the return value. Maybe len="RETURN()"? -->
```
NogginBops, line 769: @SunSerega what do you think about this?
Should we introduce `RETURN()` or should we just leave this unhandled?

SunSerega: I wouldn't mind either way, but also while this is the only case - it doesn't matter much, since this case would need to be handled manually anyway, and IMO `len` attribute values are not yet consistent enough to use as notification for consumers, like "some new, yet unexpected value appeared, need to figure out how to handle this". Mainly because of all the `COMPSIZE`-s.

NogginBops: Fair enough, I could remove this comment or leave it in. 

I'll leave it to @oddhack to say whether this comments gets to stay or gets removed.

Issue #572, opened 2023-04-02
XID type is missing from glx.xml


codingatc:
The other types coming from X11 are present such as `Display`. At minimum I think `XID` should be added to the xml as a placeholder like the other types. I am happy to add a PR if there is agreement on the fix.

oddhack:
Sure, that's fine. Just add it to the dummy placeholder section like the others (as long as you're there, it should say "X / OpenGL types", not just "X types" in the comment, for completeness).

PR #573, opened 2023-04-03
Update glx.xml to include the XID type.

codingatc:
Adds the missing `XID` type for the X headers as a placeholder type, in the same manner other types from X such as `Display` are. In addition, the comment is updated to make it clearer.

See: https://github.com/KhronosGroup/OpenGL-Registry/issues/572


oddhack:
LGTM, thanks!

PR #576, opened 2023-05-19
[XML] Performance query group and class definitions

SunSerega:


pdaniell-nv:
@Perksey any objections to this change from you?

Perksey (approved):
Sorry, thanks for the ping. Seems simple enough, no objections :)

NogginBops:
It might be good to do something like we did with `kinds` where we keep a list of the ones that are present in the specification. Otherwise I have no problem with this.

NogginBops:
I guess I mostly wanted some way to know about what a "perf query" is as it's an extension concept and not part of the core OpenGL specification. But maybe gl.xml isn't the right place for that.

pdaniell-nv:
This is approved to merge if @NogginBops doesn't object.

SunSerega:
> I guess I mostly wanted some way to know about what a "perf query" is

Yeah, that's understandable...
But I think starting from that is backward. For the use cases I see - I expect people to first understand a function, which uses a class. Then, while looking at another function - notice the handle of the same class is needed, so they must be connected. And that alone would already give the consumer all the info there is about the said class.

NogginBops:
Sorry for responding so late! I'm absolutely fine with merging this PR.

pdaniell-nv:

PR #580, opened 2023-06-23
Separate `PathFontStyle` into its own `<enums>` block

SunSerega:
Also removed the `NV` suffix to match the `group=` attributes...


pdaniell-nv:
@oddhack, this is approved to merge if you are happy with it.

PR #588, opened 2023-09-15
Add QCOM_texture_lod_bias

tate-hornbeck:
This extension reintroduces LOD bias
as an API state concept. It is based
off the existing 1.1 extension
EXT_texture_lod_bias.


oddhack:
@pdaniell-nv @tate-hornbeck this needs the corresponding changes to xml/gl.xml to define the extension APIs.

PR #592, opened 2023-10-21
Removed "len" XML attribute from "pointer" parameter declarations of "glVertexAttribIPointer", "glVertexAttribLPointer" and "glVertexAttribPointer" commands

Rytis-Stan:
The removal was done because the "pointer" parameter is not mean to act as an array, but simply as a relative integer offset when defining vertex attributes

Looks like the "len" attribute was added by accident

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -28200,7 +28200,7 @@ typedef unsigned int GLhandleARB;
             <param><ptype>GLint</ptype> <name>size</name></param>
             <param group="VertexAttribIType"><ptype>GLenum</ptype> <name>type</name></param>
             <param><ptype>GLsizei</ptype> <name>stride</name></param>
-            <param len="COMPSIZE(size,type,stride)">const void *<name>pointer</name></param>
+            <param>const void *<name>pointer</name></param>
         </command>
         <command>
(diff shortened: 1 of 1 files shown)


Perksey (approved):
Good point, the data it's pointing to is of no relevance (because it is not a memory pointer in essence) and therefore we should not be specifying its size.

NogginBops:
This is likely a leftover from the client-side vertex arrays where these were actual pointers to vertex date.
https://www.khronos.org/opengl/wiki/Client-Side_Vertex_Arrays

I'm fine with removing the length param here tough as that API usage is very outdated.

Rytis-Stan:
> This is likely a leftover from the client-side vertex arrays where these were actual pointers to vertex data. https://www.khronos.org/opengl/wiki/Client-Side_Vertex_Arrays
> 
> I'm fine with removing the length param here tough as that API usage is very outdated.

Thanks for the link!

pdaniell-nv:
@oddhack, if you're happy with this change it is approved to merge.

oddhack (approved):
Seems OK. I don't think many people are writing pre-VBO applications in compatibility mode, these days.

PR #598, opened 2023-11-23
Removed "len" XML attribute from "pointer" parameter declarations of "glVertexAttribIPointerEXT", "glVertexAttribLPointerEXT" and "glVertexAttribPointerARB" commands

Rytis-Stan:
The removal was done because the "pointer" parameter is not mean to act as an array, but simply as a relative integer offset when defining vertex attributes. An equivalent fix was done previously for the related commands: "glVertexAttribIPointer", "glVertexAttribLPointer" and "glVertexAttribPointer" (related PR: https://github.com/KhronosGroup/OpenGL-Registry/pull/592)

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -28208,7 +28208,7 @@ typedef unsigned int GLhandleARB;
             <param><ptype>GLint</ptype> <name>size</name></param>
             <param group="VertexAttribIType"><ptype>GLenum</ptype> <name>type</name></param>
             <param><ptype>GLsizei</ptype> <name>stride</name></param>
-            <param len="COMPSIZE(size,type,stride)">const void *<name>pointer</name></param>
+            <param>const void *<name>pointer</name></param>
             <alias name="glVertexAttribIPointer"/>
         </command>
         <command>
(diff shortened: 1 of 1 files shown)


NogginBops:
I don't have much of an opinion on this, but it does seem like `glVertexAttribIPointerEXT` should actually just take a pointer to an array of indices directly if I'm reading the specification right as the `EXT_gpu_shader4` is written on top of OpenGL 2.0.

Not sure what is expected of a driver that implements these methods, it would make sense they they would follow modern `glVertexAttribPointer*` conventions but I'm not sure if this is what the spec really says...

But I guess `glVertexAttribPointerARB` from either `GL_ARB_vertex_shader` or `GL_ARB_vertex_program` should behave identically to `glVertexAttribPointer` in modern drivers so I guess that is how it's supposed to be?

Maybe someone from Khronos can clarify the intended behavior for these functions.

pdaniell-nv:
The "EXT" variants of these functions were all promoted to the non-EXT core versions, so I think all drivers will treat them as aliases and therefore implement them in exactly the same way.

pdaniell-nv:
In most cases yes promoted functions become just aliases in the implementation. There are some cases where there is a small divergence between the extension function and the core function, due to spec differences in the function behavior, in which case typically a subset of the function implementation is common. In this case they are aliases.

oddhack:
We don't actually tag aliases in gl.xml (unlike the Vulkan XML). There are several possible interpretations of 'alias' for GL commands:

* Client alias in libGL.so - calling the two functions goes to the same client code via thunks or loader features or whatever.
* GLX protocol alias - different entry points in libGL.so that create the same GLX protocol which lands in the same place in the driver (and the same thing in direct rendering).
* Different client *and* driver entry points, which might or might not have somewhat different behavior (hopefully called out in the spec or extension document, if so).

pdaniell-nv:

PR #600, opened 2024-02-03
Put some implementation dependent values of FSR under pipeline FSR instead of attachment FSR

PanGao-h:
In Vulkan, `fragmentShadingRateNonTrivialCombinerOps` must be VK_FALSE unless either the `primitiveFragmentShadingRate` or `attachmentFragmentShadingRate` feature is supported. But the table here is under shading_rate_attachment alone. It's a bit confusion.  This MR tries to make the connection between Vulkan/OpenGL ES explicit.

### DIFF
--- a/extensions/EXT/EXT_fragment_shading_rate.txt
+++ b/extensions/EXT/EXT_fragment_shading_rate.txt
@@ -633,16 +633,16 @@
                                                                                                   each texel in a fragment shading
                                                                                                   rate attachment
     MAX_FRAGMENT_SHADING_RATE_ATTACHMENT_LAYERS_EXT             Z+    GetIntegerv  1              maximum number of layers in a
-                                                                                                  fragment shading rate attachment    13.X.3     
+                                                                                                  fragment shading rate attachment    13.X.3
+    FRAGMENT_SHADING_RATE_ATTACHMENT_WITH_-                     B    GetBooleanv  -               support for attachment shading      13.X.3
+    DEFAULT_FRAMEBUFFER_SUPPORTED_EXT                                                             rates on the default framebuffer
+[[If GL_EXT_fragment_shading_rate is supported]]
     FRAGMENT_SHADING_RATE_WITH_SHADER_DEPTH_STENCIL_WRITES_SUPPORTED_EXT
                                                                 B     GetBooleanv  -              support for writing FragDepth from  13.X.4
                                                                                                   a fragment shader for multi-pixel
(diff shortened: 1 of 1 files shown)


PanGao-h:
In a detailed check with Vulkan spec for [Limite Requirements](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap48.html#limits-minmax), move more items that are covered in feature [pipelineFragmentShadingRate](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap47.html#features-pipelineFragmentShadingRate). For example: 

|Type|Limit|Feature|
|--|--|--|
|VkBool32 | fragmentShadingRateWithShaderDepthStencilWrites | pipelineFragmentShadingRate|
|VkBool32 | fragmentShadingRateWithSampleMask | pipelineFragmentShadingRate|

extensions/EXT/EXT_fragment_shading_rate.txt
```diff
@@ -633,15 +633,16 @@
                                                                                                   each texel in a fragment shading
                                                                                                   rate attachment
     MAX_FRAGMENT_SHADING_RATE_ATTACHMENT_LAYERS_EXT             Z+    GetIntegerv  1              maximum number of layers in a
-                                                                                                  fragment shading rate attachment    13.X.3     
+                                                                                                  fragment shading rate attachment    13.X.3
+    FRAGMENT_SHADING_RATE_ATTACHMENT_WITH_-                     B    GetBooleanv  -               support for attachment shading      13.X.3
+    DEFAULT_FRAMEBUFFER_SUPPORTED_EXT                                                             rates on the default framebuffer
+[[If GL_EXT_fragment_shading_rate is supported]]
```
janharaldfredriksen-arm, line 639: Maybe also move the 
>[[If GL_EXT_fragment_shading_rate_attachment is supported]]

statement from line 609 to line 614 - so it groups all the attachment states?

PanGao-h: Nice. And the patch is updated accordingly.

janharaldfredriksen-arm:
I think this got fixed https://github.com/KhronosGroup/OpenGL-Registry/pull/618

(I don't remember why I created a different pull request when this already existed.)

pdaniell-nv:
> I think this got fixed #618
> 
> (I don't remember why I created a different pull request when this already existed.)

Based on this I'll close this. @PanGao-h please reopen if you disagree.

PR #603, opened 2024-02-14
Update EXT_texture_format_BGRA8888.txt

kusma:
As agreed on today's call, let's make GL_BGRA8_EXT an allowed internalformat as well.

### DIFF
--- a/extensions/EXT/EXT_texture_format_BGRA8888.txt
+++ b/extensions/EXT/EXT_texture_format_BGRA8888.txt
@@ -140,6 +140,30 @@ Interactions with the OpenGL ES 2.0 specification
     --------------- ----------        ---- ---- ---- ---- ---- ----
     BGRA_EXT        color-renderable  8    8    8    8
 
+Interactions with the OpenGL ES 3.0 specification
+
+    Add the following entries to Table 3.2: Valid combinations of format,
+    type, and sized internalformat:
+
+    Format    Type           External Bytes per Pixel Internal Format
+    ------    ----           ------------------------ ---------------
+    BGRA_EXT  UNSIGNED_BYTE  4                        BGRA_EXT, BGRA8_EXT
+
+
+    Modify Section 3.8.3 (Texture Image Specification), p 128
+
+    (add the following required Texture and renderbuffer color formats, p 128)
+
+    - BGRA_EXT
+
+    (add the following to Table 3.13: "Correspondence of sized internal color
+    formats to base internal formats...", beginning on p 130)
+
+    Sized           Base            R    G    B    A    Shared Color-     Texture-
+    Internal Format Internal Format bits bits bits bits bits   renderable filterable
+    --------------- --------------- ---- ---- ---- ---- ------ ---------- ----------
+    BGRA_EXT        BGRA_EXT        8    8    8    8           x          x
+
 Revision History
 
     0.1,  26/04/2005  sks:  Initial revision.
--- a/extensions/EXT/EXT_texture_format_BGRA8888.txt
+++ b/extensions/EXT/EXT_texture_format_BGRA8888.txt
@@ -20,7 +20,7 @@ Status
 
 Version
 
-    1.3, 12 September 2016
+    1.4, 21 February 2024
 
 Number
 
@@ -58,6 +58,7 @@ Overview
     Internal Format     External Format Type                    Bytes per Pixel
     ---------------     --------------- ----                    ---------------
     BGRA_EXT            BGRA_EXT        UNSIGNED_BYTE           4
+    BGRA8_EXT           BGRA_EXT        UNSIGNED_BYTE           4
     RGBA                RGBA            UNSIGNED_BYTE           4
     RGB                 RGB             UNSIGNED_BYTE           3
     RGBA                RGBA            UNSIGNED_SHORT_4_4_4_4  2
@@ -87,6 +88,11 @@ New Tokens
     and the <format> parameter of TexSubImage2D:
 
         GL_BGRA_EXT                                     0x80E1
+
+    Accepted by the <internalformat> parameter of TexImage2D, unless using
+    OpenGL ES versions prior to 3.0:
+
+        GL_BGRA8_EXT                                    0x93A1
 
 Additions to Chapter 2 of the OpenGL 1.3 Specification (OpenGL Operation)
 
@@ -154,7 +160,7 @@ Interactions with the OpenGL ES 3.0 specification
 
     (add the following required Texture and renderbuffer color formats, p 128)
 
-    - BGRA_EXT
+    - BGRA_EXT and BGRA8_EXT
 
     (add the following to Table 3.13: "Correspondence of sized internal color
     formats to base internal formats...", beginning on p 130)
@@ -163,6 +169,7 @@ Interactions with the OpenGL ES 3.0 specification
     Internal Format Internal Format bits bits bits bits bits   renderable filterable
     --------------- --------------- ---- ---- ---- ---- ------ ---------- ----------
     BGRA_EXT        BGRA_EXT        8    8    8    8           x          x
+    BGRA8_EXT       BGRA_EXT        8    8    8    8           x          x
 
 Revision History
 
@@ -173,3 +180,4 @@ Revision History
     1.2,  26/10/2009  Benj Lipchak: add EXT suffix to BGRA token.
     1.3,  12/09/2016  Tobias Hector: Added interaction with ES 2.0 (made it renderable).
                       Also made revision dates use the same (dd/mm/yyyy) format.
+    1.4,  14/02/2024  Erik Faye-Lund: Add GL_BGRA8_EXT for ES 3.0.
--- a/extensions/EXT/EXT_texture_format_BGRA8888.txt
+++ b/extensions/EXT/EXT_texture_format_BGRA8888.txt
@@ -43,14 +43,14 @@ Overview
 
     Internal Format     External Format Type                    Bytes per Pixel
     ---------------     --------------- ----                    ---------------
-    RGBA                RGBA             UNSIGNED_BYTE          4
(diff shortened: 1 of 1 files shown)


pdaniell-nv (approved):

zmike (approved):
LGTM (with version updated)

SunSerega:
What about the XML?
I guess the only change is that this:
```
        <enum value="0x93A1" name="GL_BGRA8_EXT"/>
```
Should be:
```
        <enum value="0x93A1" name="GL_BGRA8_EXT" group="PixelFormat"/>
```

Perksey:
@SunSerega you are right that it should be part of the PixelFormat group, but this fact is not changed by this PR and therefore is effectively out of scope. What would be in scope for this PR is adding the InternalFormat group as well, but given that neither group is present currently I suggest one of the XML editors fix this in a subsequent PR.

kusma:
By the way, I think I've come to the realization that the change in this MR needs to actually be an interaction with OpenGL ES 3.0 and Desktop OpenGL, because in OpenGL ES 1.x and 2.x it's not really legal to use different formats and internalformats... This basically depends on sized internalformat support.

kusma:
> By the way, I think I've come to the realization that the change in this MR needs to actually be an interaction with OpenGL ES 3.0 and Desktop OpenGL, because in OpenGL ES 1.x and 2.x it's not really legal to use different formats and internalformats... This basically depends on sized internalformat support.

Actually, GLES 2 has *some* support for sized internalformats... But not to the full extent, so it seems better to limit this to OpenGL ES 3.0 and later... That's also where we see the problems in current implementations.

extensions/EXT/EXT_texture_format_BGRA8888.txt
```diff
@@ -135,12 +141,35 @@ Interactions with the OpenGL ES 2.0 specification
...
-    |-----------------|------------------|------|------|------|------|------|------|
-    | Sized           | Renderable       | R    | G    | B    | A    | D    | S    |
-    | Internal Format | Type             | bits | bits | bits | bits | bits | bits |
-    |-----------------|------------------|------|------|------|------|------|------|
-    | GL_BGRA_EXT     | color-renderable | 8    | 8    | 8    | 8    |      |      |
-    |-----------------|------------------|------|------|------|------|------|------|
+    Sized           Renderable        R    G    B    A    D    S
+    Internal Format Type              bits bits bits bits bits bits
+    --------------- ----------        ---- ---- ---- ---- ---- ----
+    BGRA_EXT        color-renderable  8    8    8    8
```
kusma, line 147: @fooishbar pointed out to me elsewhere that it would be *really* convenient if `BGRA8_EXT` was still added here, because that makes things more consistent for `glRenderbufferStorage()`, which makes format handling in Weston a lot cleaner. So I think I'll add that as well, if people don't mind. After all, that's what I did at first, so I assume people are happy with it...

fooishbar: Yeah, I asked for it because it seemed very incongruous (and a real footgun) that BGRA_EXT would be an acceptable sized internal format but BGRA8_EXT would not.

kusma:
CTS tests added here: https://github.com/KhronosGroup/VK-GL-CTS/pull/449

extensions/EXT/EXT_texture_format_BGRA8888.txt
```diff
@@ -135,12 +141,35 @@ Interactions with the OpenGL ES 2.0 specification
...
+    Format    Type           External Bytes per Pixel Internal Format
+    ------    ----           ------------------------ ---------------
+    BGRA_EXT  UNSIGNED_BYTE  4                        BGRA_EXT, BGRA8_EXT
+
+
+    Modify Section 3.8.3 (Texture Image Specification), p 128
+
+    (add the following required Texture and renderbuffer color formats, p 216)
+
+    - BGRA_EXT
```
fooishbar, line 164: I'm not sure why BGRA8 is missing here, given that we've added it for both texture specification and renderbuffer storage?

extensions/EXT/EXT_texture_format_BGRA8888.txt
```diff
@@ -135,12 +141,35 @@ Interactions with the OpenGL ES 2.0 specification
...
+
+    (add the following to Table 3.13: "Correspondence of sized internal color
+    formats to base internal formats...", beginning on p 130)
+
+    Sized           Base            R    G    B    A    Shared Color-     Texture-
+    Internal Format Internal Format bits bits bits bits bits   renderable filterable
+    --------------- --------------- ---- ---- ---- ---- ------ ---------- ----------
+    BGRA_EXT        BGRA_EXT        8    8    8    8           x          x
```
fooishbar, line 172: Ditto

extensions/EXT/EXT_texture_format_BGRA8888.txt
kusma: Yeah, that seems like a mistake here.

Also "p 216" is wrong.

extensions/EXT/EXT_texture_format_BGRA8888.txt
kusma: Yeah, seems like a mistake as well.

extensions/EXT/EXT_texture_format_BGRA8888.txt
fooishbar: Heh yeah, I was wondering about p216, but I figured it was a reference to something that indirectly backlinks to this table, noting the required renderbuffer storage formats.

kusma: Looking at this, I think I just kinda half-did things, and probably lost a bit of track of what I was doing. This entire section seems to just be about adding an explicit interaction with ES3, because we intend to diverge from ES2 a bit there.

So this should probably be its own preceding commit, and then the "real" sized enums should be added in this commit.

extensions/EXT/EXT_texture_format_BGRA8888.txt
kusma: Fixed

extensions/EXT/EXT_texture_format_BGRA8888.txt
kusma: > This entire section seems to just be about adding an explicit interaction with ES3, simply to make it clearer.

Strike that last part, the reason is because we wanted to only change this for ES3 and later...

extensions/EXT/EXT_texture_format_BGRA8888.txt
```diff
@@ -135,12 +141,35 @@ Interactions with the OpenGL ES 2.0 specification
...
+    Format    Type           External Bytes per Pixel Internal Format
+    ------    ----           ------------------------ ---------------
+    BGRA_EXT  UNSIGNED_BYTE  4                        BGRA_EXT, BGRA8_EXT
+
+
+    Modify Section 3.8.3 (Texture Image Specification), p 128
+
+    (add the following required Texture and renderbuffer color formats, p 128)
+
+    - BGRA_EXT and BGRA_EXT
```
fooishbar, line 163: BGRA_EXT and BGRA_EXT? :)

fooishbar:
lgtm, thankyou!

extensions/EXT/EXT_texture_format_BGRA8888.txt
oddhack: I generally say something like "The nnn-th paragraph of section S.P.Q" in an attempt to avoid hardwiring layout dependencies. It at least is likely to work for that specific version of the spec it's written against but later point updates may defeat it. Additional context is also sometimes helpful.

extensions/EXT/EXT_texture_format_BGRA8888.txt
```diff
@@ -88,6 +89,11 @@ New Tokens
 
         GL_BGRA_EXT                                     0x80E1
 
+    Accepted by the <internalformat> parameter of TexImage2D, unless using
+    OpenGL ES versions prior to 3.0:
```
kusma, line 93: The more I think about it, the more I think that limiting this to ES 3 is a mistake. I think when I proposed this, I thought the sized internal format support for ES 2 was more limited than it is. But even if it is, it means we need to complicate implementations and the CTS tests for little reason here.

Thoughts? Should this just apply to all versions instead?

zmike:
WG has approved the latest version.

extensions/EXT/EXT_texture_format_BGRA8888.txt
zmike: Supporting all versions has been approved by the WG.

kusma: OK. Added a commit to include ES2. I don't think changing ES1 is a good idea, sorry for not being clear about that earlier.

kusma: I also decided to keep the interaction with the ES3 spec, even if it could be removed now. It seems clearer this way.

kusma: Looking at [the minutes](https://gitlab.khronos.org/opengl/API/-/wikis/OpenGL%7CES-Meetings/2024-05-22-minutes), it seems like adding ES2 is what was approved, so I assume this makes the approval stand?

zmike: That's correct. I meant that supporting ES2 was approved in my previous statement, not that ES1 is supported.

oddhack:
> WG has approved the latest version.

@zmike could you set the Signed Off milestone if that's what this means?

zmike:
> > WG has approved the latest version.
> 
> @zmike could you set the Signed Off milestone if that's what this means?

The comment was for the author. I'll set the milestone when this is ready to merge.

kusma:
The CTS tests landed in the Khronos GitLab repo. Is it time to merge this PR?

PR #606, opened 2024-03-08
Allocate enums for Samsung

stonesthrow:
Request a block of enums for Samsung.
Private enums for ANGLE use.


stonesthrow:
@oddhack , If I got this request right, please allocate this block to me. x9700-970F

oddhack:
LGTM.

PR #610, opened 2024-05-07
Update vendor to "QCOM" and fix group for new enums

MatAshish:
This updates vendor name from "Qualcomm" to "QCOM" and fix group for new enums 0x9710 and 0x9711.

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -6928,9 +6928,9 @@ typedef unsigned int GLhandleARB;
             <unused start="0x9690" end="0x969F" vendor="ANGLE"/>
     </enums>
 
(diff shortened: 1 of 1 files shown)


SunSerega:
It's still a bit wrong... `TextureParameter` also doesn't exist as a group - but there is `TextureParameterName`.

SunSerega:
Also, almost all enums are mentioned again below, in `<require>` block, within `<feature>` or `<extension>`.
But it's alright to not have them there, if you have reserved these enums for an extension, which you are planning to publish later.

MatAshish:
I will fix it with the new extension PR soon.

PR #618, opened 2024-07-03
GL_EXT_fragment_shading_rate: clarify state table

janharaldfredriksen-arm:
For GL_EXT_fragment_shading_rate, someone asked me if all the state in Implementation Dependent Values was gated on support for attachment shading rates (GL_EXT_fragment_shading_rate_attachment). That's not the intent, but the state table can definitely be read that way. (This was probably an 'accident' of us adding more state during the spec process.)

I've tried to clarify this by moving some lines around - where the only actual change is that the "[[If GL_EXT_fragment_shading_rate_attachment is supported]]" statement is now in the middle of table.

### DIFF
--- a/extensions/EXT/EXT_fragment_shading_rate.txt
+++ b/extensions/EXT/EXT_fragment_shading_rate.txt
@@ -868,4 +868,5 @@
         ----  --------   --------  ----------------------------------------------
          1    2022-05-30 jhf       Initial EXT draft
          2    2022-07-07 jhf       Fixed typos in the spec text
-         3    2022-07-03 jhf       Clarify Implementation Dependent Values table
+         3    2024-07-03 jhf       Clarify Implementation Dependent Values table
+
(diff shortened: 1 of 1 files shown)


extensions/EXT/EXT_fragment_shading_rate.txt
```diff
@@ -867,3 +868,4 @@
         ----  --------   --------  ----------------------------------------------
          1    2022-05-30 jhf       Initial EXT draft
          2    2022-07-07 jhf       Fixed typos in the spec text
+         3    2022-07-03 jhf       Clarify Implementation Dependent Values table
```
janharaldfredriksen-arm, line 871:
```suggestion
         3    2024-07-03 jhf       Clarify Implementation Dependent Values table

```

zmike (approved):

oddhack (approved):
LGTM

PR #621, opened 2024-07-16
Remove the last of the old kind attributes

NogginBops:
This PR fixes the last stuff that was left in in #534. Basically this PR replaces and removes the last `CheckedFloat` and `CheckedInt32` kinds (as well a few others kinds that where leftovers and never described formally).

I thought I had fixed these things but it turned out that I had some +1 year old commits in this branch that I never opened a PR for.

cc @SunSerega

### DIFF
--- a/xml/gl.xml
+++ b/xml/gl.xml
@@ -8313,45 +8313,45 @@ typedef unsigned int GLhandleARB;
         </command>
         <command>
             <proto>void <name>glClearColor</name></proto>
-            <param kind="Color"><ptype>GLfloat</ptype> <name>red</name></param>
-            <param kind="Color"><ptype>GLfloat</ptype> <name>green</name></param>
-            <param kind="Color"><ptype>GLfloat</ptype> <name>blue</name></param>
-            <param kind="Color"><ptype>GLfloat</ptype> <name>alpha</name></param>
+            <param kind="Clamp01,Color"><ptype>GLfloat</ptype> <name>red</name></param>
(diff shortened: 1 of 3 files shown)


NogginBops:
I think I might have messed up the git stuff as I think some of these commits are part of #534. Not sure if I can fix this in any good ways.

SunSerega:
I think this is because #534 was squash-merged.
In other words, an equivalent commit was added to the `main` branch, so this PR is compatible, but the commits of #534 themselves are not on main - so they are shown in changes.

One way to fix this is to rebase the `more-kind-changes` branch onto the main branch.
Rebase would make the current `main` - the origin of the changes, as if you started from scratch and cherry-picked every commit of this PR in the same order, discarding the now no-op commits (because they were already merged into main).

Though rebase might be messy in this case.
It might be easier to `reset --hard` the `more-kind-changes` branch onto the main branch, and manually cherry-pick the only relevant commits from the remote version of this branch.

Either way, you'd need to `push --force` (or `push --force-with-lease`) to fix how it looks on GitHub.
Before forcing new history onto GitHub, I recommend that you diff the remote and the local (rebased) states of the `more-kind-changes` branch to make sure there aren't any changes you were unaware of.

NogginBops:
#622 fixes the git history. Used `git cherry-pick` in the specific commits that where missing in #534 into a new branch based on `main`.

SunSerega:
There are 3 changes in gl.xml between the last 2 PRs.
Well, the first 2 look logical, but the third one is a `kind="Range[0; x]"` => `kind="CheckedInt32"` change.

NogginBops:
Will fix!

PR #622, opened 2024-07-16
Remove the last of the old kind attributes

NogginBops:
This PR fixes the last stuff that was left in in #534. Basically this PR replaces and removes the last `CheckedFloat` and `CheckedInt32` kinds (as well a few others kinds that where leftovers and never described formally).

I thought I had fixed these things but it turned out that I had some +1 year old commits in this branch that I never opened a PR for.

cc @SunSerega 

This is a fixed version of #621 with cleaner git history.


SunSerega (approved):
Sorry, took me a while to get back to this, partially because there were some conflicts with my state of XML.
But now that I looked at them, I found that those conflicts were mistakes on my part, and changes from this PR make more sense.

And everything else here also LGTM.

pdaniell-nv:
@oddhack can you review this PR, and if you're happy with it, it's approved for merge.

PR #623, opened 2024-07-16
Add len attribute to glGetObjectLabelKHR

NogginBops:
This was an old commit I had on one of my branches. It adds a `len="1"` attribute to glGetObjectLabelKHRs length parameter which is a return parameter of length 1.


pdaniell-nv (approved):
OpenGL/ES WG approved

pdaniell-nv:
@oddhack this fix can be merged.

PR #626, opened 2024-07-20
Reserve 0x9720-0x973F for ANGLE

ShabbyX:


NogginBops:
The comment `"github pull request"` is pretty non-descriptive... If nothing else shouldn't the comment mention which pull request?

xml/gl.xml
```diff
@@ -7023,6 +7023,10 @@ typedef unsigned int GLhandleARB;
             <unused start="0x9712" end="0x971F" vendor="QCOM"/>
     </enums>
 
+    <enums namespace="GL" start="0x9720" end="0x973F" vendor="ANGLE" comment="github pull request">
```
oddhack, line 7026:
```suggestion
    <enums namespace="GL" start="0x9720" end="0x973F" vendor="ANGLE" comment="github pull request 626">
```

oddhack:
> The comment `"github pull request"` is pretty non-descriptive... If nothing else shouldn't the comment mention which pull request?

Yes, or nothing. In the past we put Bugzilla numbers in there but now people can just backtrack the git blame history to find this commit if they're concerned.

xml/gl.xml
oddhack: @ShabbyX suggested edit per the other comment here but this is fine.

For some reason github is not letting me apply the suggestion so you will have to do it, then I'll merge this.

oddhack: Maybe it won't let me make a change since this is sourced from your repo clone? N.b. this is one reason I encourage Khronos people to just create branches in the target repo.

SunSerega: There is a checkmark to allow maintainers to edit PRs (by adding commits).
Also, I hadn't tested it fully, but in some cases `git push`-s to another repo when these edits are allowed were working only with `--force`, and at the end it looked like it wasn't a force-push. But that sounds very wrong and undertested.

oddhack: I gave myself Maintainer and Admin rights and still getting the same behavior, so it seems more likely to be that it's sourced from @ShabbyX's own clone.

Rather than futz around with it more, am going to merge and fix the comment afterwards.

ShabbyX: I did actually try to push to a branch in this repo directly, but it wouldn't let me. I checked the terminal I had open just now and realized `origin` was the Chromium mirror as I did this in a third_party directory in ANGLE, my bad.

oddhack: You should have team access now.

PR #629, opened 2024-07-28
Add T4F_C4UB_N3F_V4F and T4F_C4F_N4F_V4F formats

jopadan:
Proposal to add glInterleavedArrays vertex array standard formats:

- T4F_C4UB_N3F_V4F  12 byte 4x3 matrix
- T4F_C4F_N4F_V4F  aligned 16 byte 4x4 square matrix

See: [mesa merge request](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/30397) too


oddhack:
To add new formats would require creating an extension and getting vendors to implement that extension in their drivers. If you can convince them to do that you're welcome to, but this PR literally accomplishes nothing other than adding a few lines to the XML.

Issue #617, opened 2024-06-27
GL_GLEXT_VERSION missing in ES2 and ES3 headers


emersion:
`api/GL/glext.h` exposes `GL_GLEXT_VERSION`. However, this symbol is missing from GLES2/GLES3 headers. This makes it difficult to check for the version of the headers programmatically (e.g. to ensure the headers are new enough to contain required extensions).

zmike:
WG 2024-07-17: Adding to ES headers is approved. Please create a pull request.

PR #630, opened 2024-07-30
Add version define to gl2ext.h

emersion:
Closes: https://github.com/KhronosGroup/OpenGL-Registry/issues/617

* * *

- Should we use a different symbol name, in case users include both `glext.h` and `gl2ext.h` from different releases?
- Should I re-generate the headers, or should I leave that to the person in charge of the next release?

### DIFF
--- a/xml/genglvnd.py
+++ b/xml/genglvnd.py
@@ -213,7 +213,7 @@ genDateCommentString = [
     ''
 ]
 
-# GL_GLEXT_VERSION is defined only in glext.h
+# GL_GLEXT_VERSION is defined only in glext.h and gl2ext.h
 glextVersionStrings = [
     format("#define GL_GLEXT_VERSION %s" % time.strftime("%Y%m%d")),
     ''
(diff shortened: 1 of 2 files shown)


xml/genglvnd.py
```diff
@@ -364,7 +364,7 @@ def makeREstring(list):
         defaultExtensions = 'gles2',                # Default extensions for GLES 2
         addExtensions     = None,
         removeExtensions  = None,
-        prefixText        = prefixStrings + apiEntryPrefixStrings + genDateCommentString,
+        prefixText        = prefixStrings + apiEntryPrefixStrings + genDateCommentString + glextVersionStrings,
```
zmike, line 367: `genDateCommentString` should be removed now since it was added because the version string wasn't present.

xml/genheaders.py
```diff
@@ -296,7 +296,7 @@ def makeREstring(list):
         defaultExtensions = 'gles2',                # Default extensions for GLES 2
         addExtensions     = None,
         removeExtensions  = None,
-        prefixText        = prefixStrings + apiEntryPrefixStrings + genDateCommentString,
+        prefixText        = prefixStrings + apiEntryPrefixStrings + genDateCommentString + glextVersionStrings,
```
zmike, line 299: `genDateCommentString` should be removed now since it was added because the version string wasn't present.

zmike (commented):
WG 14 August 2024: With date string removed, this is approved by the WG.

zmike:
@oddhack Once date string is removed, please review and merge if you are happy with the changes.

emersion:
Done!

oddhack:
It seems fine. I will regenerate the headers.

PR #633, opened 2024-09-23
Add len and group attributes to glx.xml

NogginBops:
Similar to #569 but for glX.

### DIFF
--- a/xml/glx.xml
+++ b/xml/glx.xml
@@ -187,18 +187,18 @@ typedef unsigned __int64 uint64_t;
     <!-- Bitmasks each have their own namespace, as do a few other
          categories of enumeration -->
 
-    <enums namespace="GLXStringName">
-        <enum value="0x1"           name="GLX_VENDOR"/>
-        <enum value="0x2"           name="GLX_VERSION"/>
-        <enum value="0x3"           name="GLX_EXTENSIONS"/>
+    <enums namespace="StringName">
+        <enum value="0x1"           name="GLX_VENDOR" group="StringName"/>
+        <enum value="0x2"           name="GLX_VERSION" group="StringName"/>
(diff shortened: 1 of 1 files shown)


xml/glx.xml
```diff
@@ -1051,11 +1051,11 @@ typedef unsigned __int64 uint64_t;
         </command>
         <command>
             <proto><ptype>__GLXextFuncPtr</ptype> <name>glXGetProcAddressARB</name></proto>
-            <param>const <ptype>GLubyte</ptype> *<name>procName</name></param>
+            <param kind="String">const <ptype>GLubyte</ptype> *<name>procName</name></param>
         </command>
         <command>
             <proto><ptype>__GLXextFuncPtr</ptype> <name>glXGetProcAddress</name></proto>
-            <param>const <ptype>GLubyte</ptype> *<name>procName</name></param>
+            <param kind="String">const <ptype>GLubyte</ptype> *<name>procName</name></param>
```
SunSerega, line 1058: I don't fully remember, but didn't we agree on something like `gl::String` to reference another XML document?

NogginBops: I was under the impression that that was for enumerated groups only. I think we could possibly extend it but then I think that requires further changes to the xml schema

SunSerega: Ah, right, this is the first time the `kind` attribute appeared outside of `gl.xml`.

Well, it's only slightly bugging me that the description is in another XML file.

I'm fine with keeping kinds non-namespaced, in case you think we wouldn't need that separation, but let's agree on this now.

NogginBops: Personally I'm fine with both.
I personally think we won't need namespaces for kinds as I don't think we'll need the same name to mean different things in the different xml files.

PR #634, opened 2024-09-27
Update issue text for EXT_texture_shadow_lod

gnl21:
One issue referred to a bug that has now been fixed, so is no longer relevant, so this text is removed.

An additional issue is added describing which stages the bias variants are added to. The normative text was already correct but this may aid clarity.

Fixes internal issues 75 and 77. @lexaknyazev


zmike (approved):

pdaniell-nv:

PR #643, opened 2025-02-03
Add GL_MESA_texture_const_bandwidth

robclark:
GL_MESA_texture_const_bandwidth adds a way for an app to request the use of a format which does have data dependent bandwidth characteristics, in order to avoid a potential side-channel.

### DIFF
--- a/api/GL/glext.h
+++ b/api/GL/glext.h
@@ -32,7 +32,7 @@ extern "C" {
 #define GLAPI extern
 #endif
 
-#define GL_GLEXT_VERSION 20250129
+#define GL_GLEXT_VERSION 20250203
 
 #include <KHR/khrplatform.h>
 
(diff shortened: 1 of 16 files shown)


robclark:
fwiw, original MESA implementation is at https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/25464

Triang3l:
The enumeration name finally chosen in Mesa, `GL_CONST_BW_TILING_MESA`, seems somewhat unusual considering that OpenGL doesn't seem to use such shortenings frequently. `_CONSTANT` is referenced 27 times in glext.h, while `_CONST` is used only for `CONST_EYE` in the ancient `GL_NV_texture_shader`. The original `GL_CONSTANT_BANDWIDTH_TILING_MESA` suggestion from that GitLab merge request seems much more consistent with the rest of OpenGL and more intuitive.

robclark:
So what is the verdict on naming.  I can rename the enums, but renaming the extension would be an ABI break, since this extension has been in mesa for a while.

zmike:
The verdict is do whatever you want.

robclark:
> The verdict is do whatever you want.

I'd vote to land it as-is, since it has been in mesa for a while now

PR #646, opened 2025-03-26
remove GL_EXT_direct_state_access from some ES EXT extensions

zmike:
internal issue 243


extensions/EXT/EXT_buffer_storage.txt
```diff
@@ -408,11 +405,6 @@ Errors
     MAP_PERSISTENT_BIT_EXT included in <access>.
 
 
-Dependencies on GL_EXT_direct_state_access
-
-    If GL_EXT_direct_state_access is not supported, remove all references to
-    NamedBufferStorageEXT.
```
lexaknyazev, line 414: Please remove actual references to that function, starting from line 79.

lexaknyazev: Yes. This is an ES-only extension so DSA does not exist here. Note that there are 11 references to that function in the extension text.

zmike: Added another issue to document this.

lexaknyazev (approved):

nigels-com:
@zmike, @oddhack Can you clarify that `glNamedBufferStorageEXT` is definitely not relevant on desktop OpenGL with DSA?

See [nigels-com/glew Issue #454](https://github.com/nigels-com/glew/issues/454)

Niram7777:
> My reading is that it is _only_ relevant on desktop OpenGL with DSA. I believe the purpose of the PR you link to from the glew issue was to completely remove discussion of DSA interactions from ES-specific extension specs, because DSA is not supported on ES. The command is defined in the XML but will only appear as part of DSA.
> 
> But will defer to @zmike if I misunderstood something.

if DSA is not supported on ES, you will make the same PR for https://github.com/KhronosGroup/OpenGL-Registry/blob/main/extensions/EXT/EXT_external_buffer.txt ?

EDIT: I guess no because `supported="gl|gles2"`

PR #649, opened 2025-04-23
add EXT_texture_storage note about missing functionality in ES

zmike:
Internal issue 247.

### DIFF
--- a/extensions/EXT/EXT_texture_storage.txt
+++ b/extensions/EXT/EXT_texture_storage.txt
@@ -31,8 +31,8 @@ Status
 
 Version
 
-    Last Modified Date: 22 September, 2021
-    Author Revision: 29
+    Last Modified Date: 23 April, 2025
+    Author Revision: 30
 
 Number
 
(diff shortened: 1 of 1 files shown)


zmike:
@oddhack please merge

extensions/EXT/EXT_texture_storage.txt
```diff
@@ -997,8 +997,20 @@ Issues
...
     decision.
 
+    28. Should GetProcAddress return null pointers for functions not present
+    in a given version of the API?
+
+    RESOLVED. Yes, certain functions which are present in the API header may
+    return null pointers from the corresponding GetProcAddress calls. For example,
+    the gl2ext.h header contains TexStorage1DEXT and TextureStorage* references,
+    but these functionalities are not supported in OpenGL ES and will return
+    null pointers.
```
fooishbar, line 1007: I don't think this is right. GPA is defined to be independent of the currently-bound API; it can (necessarily) be used before any client API is bound, and it's defined to return a pointer which can be used on _a_ client API on that implementation, regardless of what's currently bound.

So specifying that `eglGetProcAddress("glTextureStorage2DEXT")` breaks implementations which expose both GL+DSA and GLES+TS, because they need to return the GL pointer for this query at all times. You can weaken the language to 'may', but honestly you could just delete anything referring to GPA and just note that the function prototypes were erroneously included, are not defined to have any functionality in GLES, and should not be used.

PR #664, opened 2025-10-22
Add GL_EXT_shader_realtime_clock

ashleysmithcol:
This extension was added to GLSL but not yet to the OpenGL registry.

See https://github.com/KhronosGroup/GLSL/issues/280

### DIFF
--- a/extensions/EXT/EXT_shader_realtime_clock.txt
+++ b/extensions/EXT/EXT_shader_realtime_clock.txt
@@ -1,0 +1,84 @@
+Name
+
+    EXT_shader_realtime_clock
+
+Name Strings
+
+    GL_EXT_shader_realtime_clock
+
+Contact
+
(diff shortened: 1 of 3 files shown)


ashleysmithcol:
Quick note that there are [piglit tests](https://gitlab.freedesktop.org/mesa/piglit/-/blob/main/tests/spec/ext_shader_realtime_clock/execution/clock2x32.shader_test?ref_type=heads) No precise CTS tests but the [vulkan CTSs does](https://github.com/KhronosGroup/VK-GL-CTS/blob/c2070dab0c392bd5352d9790973b920e12f7ae9c/external/vulkancts/modules/vulkan/shaderexecutor/vktShaderClockTests.cpp#L198) test the functionality which is identical.

zmike (approved):

PR #676, opened 2026-02-03
Extend the relaxation on readPixels to the multisample attachments created by core functions of  ES 3.0 and later

dreamelon:
As discussed [here](https://gerrit.khronos.org/c/vk-gl-cts/+/19265/1/modules/gles3/functional/es3fNegativeBufferApiTests.cpp#311)

> the relaxation in the EXT spec is explicitly limited to attachments created via RenderbufferStorageMultisampleEXT / FramebufferTexture2DMultisampleEXT.
> 
> So from spec perspective, implementations should still distinguish whether the multisample attachment was created by:
> 
> EXT path → relaxation applies
> Core path → no relaxation; GL_INVALID_OPERATION required

I would like to extend the relaxation to core path, would you agree @zmike @janharaldfredriksen-arm

### DIFF
--- a/extensions/EXT/EXT_multisampled_render_to_texture.txt
+++ b/extensions/EXT/EXT_multisampled_render_to_texture.txt
@@ -21,8 +21,8 @@ Status
 
 Version
 
-    Last Modified Date: June 28, 2016
-    Revision: 7
+    Last Modified Date: Feb 3, 2026
+    Revision: 8
 
 Number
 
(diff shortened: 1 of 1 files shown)


dreamelon:
Thank you @janharaldfredriksen-arm for reply. I understand your points and have found the past discussion about this issue here https://github.com/KhronosGroup/OpenGL-Registry/pull/258. RenderbufferStorageMultisampleEXT and RenderbufferStorageMultisample are not identical on OpenGL ES, I'll drop this request.

PR #680, opened 2026-04-02
Finalize MESA_map_buffer_client_pointer

Guy1524:
Changes the ABI, as #678 was accidentally merged before ready, and there are no users of this extension yet.

1) Add RELEASE_ALL flag previously discussed (https://github.com/KhronosGroup/OpenGL-Registry/pull/678#issuecomment-4059110654)

2) Require a storage flag for MAP_CLIENT_POINTER_MESA usage, so that sub-allocated buffers with backing BOs on client pointer ranges are possible without remapping.

3) Specify shared state properties of client pointer ranges. This is important since Wine is moving towards sharing all GL contexts, and the client pointer ranges shouldn't be locked to only one of these contexts.

I've done this by specifying that an incoming client pointer range is at first client state, which *can* then be consumed by MapRange, (if a mapping isn't provided for the backing BO by another thread first). But the semantics around when a range is consumed are wishy washy.

Alternatively, we could make all added ranges server side state, which would be easier to spec, but on the driver side introduces some needless code.


Guy1524:
It probably makes most sense to introduce a usage/flag to glBufferData/glBufferStorage to enable client pointer mapping for the buffer. I'll work on implementing this in Zink, where we can't remap.

Guy1524:
Mesa/Zink implementation: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/43297

Guy1524:
@zmike @marekolsak @cjacek

marekolsak:
Looks good to me.

zmike (approved):

zmike:
@oddhack please merge

PR #682, opened 2026-06-07
[XML] Remove inadvertent use of `GLvoid` in command definition

castholm:
This use of `GLvoid` introduced by #678 broke my code generator. It's a trivial fix on my end, but `GLvoid` is not used in any other command definitions and is only included as a typedef for backward compatibility, so it's probably better to use `void` everywhere for consistency.


NogginBops:
The gl.xml also explicitly has comments like `"Not an actual GL type, though used in headers in the past"` and `"No longer used in headers"`. But with this extension this type name is actually used in the headers. So I agree that I would be better to change `GLvoid` to `void` here.

zmike (approved):

